PR #25, opened 2015-01-01
Fixed typo and added a few missing codes to the list of WASAPI AUDCLNT_E_* constants

inrustwetrust:
Also verified the existing AUDCLNT_E_* and AUDCLNT_S_* constants against the Audioclient.h header from the Windows SDK (7.1A).


retep998:
Thanks!

PR #26, opened 2015-01-02
Corrected signature for GetStreamLatency (WASAPI)

inrustwetrust:
See: http://msdn.microsoft.com/en-us/library/windows/desktop/dd370874%28v=vs.85%29.aspx


retep998:
Thanks for catching this!

PR #28, opened 2015-01-04
Corrected signature for IStream::Seek

inrustwetrust:
See: http://msdn.microsoft.com/en-us/library/windows/desktop/aa380043%28v=vs.85%29.aspx


retep998:
Thanks for catching this!

PR #33, opened 2015-01-21
Add input types for SendInput

alexchandel:
These are from winuser.h, and SendInput is in User32. This complements an immediately forthcoming PR to user32-sys.


alexchandel:
@retep998 Good to merge?

PR #54, opened 2015-02-24
Expose fields of CREATESTRUCT as public

msiglreith:


retep998:
I really wish there was a lint I could enable to tell me if I accidentally made any struct fields private.

PR #58, opened 2015-03-04
User32

jojonv:
bindings added


jojonv:
AppendMenuA is missing return val

PR #59, opened 2015-03-06
Add support for d3d9.h

cmr:


cmr:
I have no idea where that failure is coming from. It builds locally...

PR #62, opened 2015-03-09
Remove common prefixes from D3D constants

cmr:
Is this a good idea? I think it's a good idea. Enums are super obnoxious otherwise. Like, 3x worse than C.


jojonv:
It's not something I care much about, but I think we should be strict about this, and adhere as much as possible to how they're defined in their respective headers. It may also be somehow confusing for someone who never used them, eg. looks for help at MSDN and then compares.

Btw. are enums best for this? There places in winapi where enum's fields are equal, does Rust allow this?

cmr:
No, it doesn't, and I strongly disagree with using Rust enums for this, but that's the decision @retep998 seems to have made.

PR #72, opened 2015-04-08
add memory map bindings

danburkert:


danburkert:
rebased against current master.

PR #73, opened 2015-04-08
QueueStatus flags should be of type UINT

danburkert:
This fixes the build on 64bit OSX.  I believe it is correct according to https://msdn.microsoft.com/en-us/library/windows/desktop/ms644940%28v=vs.85%29.aspx.


retep998:
This code should never be compiled for OS X to begin with considering it is bindings to Windows specifies libraries.

danburkert:
That's true, but Cargo's method of restricting dependencies to just windows is quite clunky, so it's helpful to have the package not fail on other platforms. Regardless, I think the type for those flags is meant to be `UINT` and not `DWORD` if I'm reading that documentation correctly.

danburkert:
rebased against current master, although you have fixed the compilation failure on non-windows systems in a better way, so I no longer really need this.  I'll leave it open in case you want to merge it anyway.  Thanks!

PR #75, opened 2015-04-08
Absorb https://github.com/bryal/dxgi-rs into dxgi-sys

bryal:
AFAIK, this adds all functions, structures, enumeraitions, constants (GUIDs), and interfaces of dxgi.
Also add some macros from dxgi-rs to winapi for more convenient construction of representations of C++ classes


retep998:
Unfortunately this completely breaks with the existing style of how things are done in this project.

PR #81, opened 2015-04-21
Handful of functions

yonran:


yonran:
The build succeeds on my Windows 8 vm, but appveyor’s gcc is giving error: `undefined reference to ‘Pie@32’`, where the [Pie function](https://msdn.microsoft.com/en-us/library/windows/desktop/dd162799%28v=vs.85%29.aspx) should be defined in gdi32.dll starting Windows 2000. Does anyone know why the error is being given? Do I have to add Pie to a def file somewhere? Is appveyor compiling on mingw or something?

retep998:
@yonran Linking is done to an import library, not to the `.dll` directly. Since Rust uses the MinGW toolchain on Windows, this means we link to MinGW's import libraries, which are known for missing things. There's not much that can really be done about it, except commenting out the failing test.

yonran:
@retep998 I have re-committed tests/gdi32.rs with Pie commented out. By the way, do you know how I could try testing with mingw32?

retep998:
Appveyor is using the 32-bit Windows nightly of Rust, with its bundled MinGW, so as long as you use the 32-bit nightly as well your test results should be the same.

PR #83, opened 2015-04-24
RI Constants

excaliburHisSheath:
This pull request changes the type of the raw input constants from `DWORD` to `USHORT`. The values these constants will be most often compared against (see the `usButtonFlags` member of [RAWMOUSE](https://msdn.microsoft.com/en-us/library/windows/desktop/ms645578%28v=vs.85%29.aspx)) have the type `USHORT` so this change saves unnecessary type casting in client code.


excaliburHisSheath:
Whoops, I didn't mean to include the XInput stuff in this PR, I'll close it and reopen with just the flag changes.

PR #87, opened 2015-04-30
added GetStartupInfoW

yonran:
Is it intentional for winapi/lib.rs to export all the modules (`pub mod winnt;`) *and* all the items in the modules (`pub use winnt::*`)? In this commit, I am exporting the new struct as `winapi::STARTUPINFOW` but not as `winapi::processthreadsapi::STARTUPINFOW`, unless you’d prefer both.


retep998:
It is intentional because it means in the docs everything will be split up by header making it easier to browse and find things, while in code you will still be able to refer to them through just `winapi::` for convenience.

yonran:
Thank you for explanation. I have modified the commit to make it `pub use processthreadsapi::*;`

yonran:
Rebased and formatted to <=99 columns.

PR #90, opened 2015-05-07
Add partial visual styles support

rpjohnst:
This adds bindings to `InitCommonControlsEx` as well as `gdi32`'s font-setting functions. This by itself isn't enough for an application to easily enable visual styles; for that it still needs to add a manifest declaring support for comctl32.dll version 6.

I've played around with linking a manifest into a Rust crate, but that's sort of outside the scope of this library and is still really hacky right now.


yonran:
Should LPINITCOMMONCONTROLSEX be a `*mut` instead of `*const`? I think that would be more consistent with other parts of this library.

Should we not `pub use commctrl` within src/lib.rs? CommCtrl.h is not included when you `#include <Windows.h>`, unlike e.g. WinUser.h.

rpjohnst:
I did `pub use commctrl` because other non-`windows.h` headers are also included, such as `d3d9`. I can change `LPINITCOMMONCONTROLSEX`, though- the only use of `InitCommonControlsEx` is a direct `*const` pointer to it as an argument to `InitCommonControlsEx()`.

PR #92, opened 2015-05-09
ktmw32-sys (Kernel Transaction Manager)

gentoo90:
* add `(Create|Commit|Rollback)Transaction` functions of `ktmw32.dll`
* add `Reg(Create|Open|Delete)Transacted` functions of `advapi32.dll`
* fixed type of `lpClass` in signatures of `RegCreateKeyExA` and `RegCreateKeyExW` in accordance with [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724844%28v=vs.85%29.aspx)


gentoo90:
Rust distribution doesn't include `ktmw32.a`, so linking to `ktmw32.dll` requires full mingw toolchain.
Do you have any plans of including `*.a` files into corresponding `*-sys` crates?

Cargo.toml
```diff
@@ -12,33 +12,15 @@ license = "MIT"
...
+d3d9-sys = { path = "lib/d3d9-sys" }
+gdi32-sys = { path = "lib/gdi32-sys" }
+kernel32-sys = { path = "lib/kernel32-sys" }
+ktmw32-sys = { path = "lib/ktmw32-sys" }
+ole32-sys = { path = "lib/ole32-sys" }
+opengl32-sys = { path = "lib/opengl32-sys" }
+shell32-sys = { path = "lib/shell32-sys" }
+user32-sys = { path = "lib/user32-sys" }
+uuid-sys = { path = "lib/uuid-sys" }
+winmm-sys = { path = "lib/winmm-sys" }
```
retep998: I need the `version = "*"` bit so that I can upload it to crates.io.

retep998:
Not relying on locally installed libraries is something I am considering.

gentoo90:
`version = "*"` added

gentoo90:
Rebased on the latest `master` and added `libktmw32.a` files (in the same way as for `dwmapi` in 019eb25d)

lib/ktmw32-sys/build.rs
```diff
@@ -1,5 +1,19 @@
 // Copyright © 2015, Peter Atashian
 // Licensed under the MIT License <LICENSE.md>
+use std::env::var;
+
+fn link_win32_lib(libname: &str) {
```
gentoo90: This should probably be common for all `*-sys` crates, but that would require another build dependency.

gentoo90:
Rebased again, and used `winapi-build` crate

PR #94, opened 2015-05-15
Add DwmEnableBlurBehindWindow function

tomaka:


tomaka:
Getting linking error to `dwmapi` when trying to use it, as it's not bundled with Rust's mingw :(

PR #97, opened 2015-05-17
Correct return types for numerous D3D9 interface methods.

iliekturtles:
Many methods including IUnknown::AddRef/Release were returning
`*mut IDirect3D...` pointers. Method arguments formatted to be one per line.

I used `rustc -Z unstable-options --pretty expanded src/d3d9.rs` and compared the output of master vs d3d9 to verify the formatting didn't catch anything unexpected.


retep998:
I prefer having the function arguments packed to multiple per line so that it takes up less space.

iliekturtles:
I will repack and re-submit.

Issue #17, opened 2014-12-02
Split code into modules


retep998:
Currently `lib.rs` is a massive file which makes editing difficult, and can bring certain editors to a crawl.

* [x] Move code from each header into separate module
* [x] Split libraries into separate crates
* [ ] Get rid of ugly `::` everywhere as soon as glob imports aren't broken.

retep998:
A design plan has been figured out which doesn't depend on recursive imports. However once recursive imports work correctly the code can be cleaned up.

retep998:
Glob imports will be working correctly in Rust 1.4, so hopefully next year I can begin the transition.

Eh2406:
What are the next steps, now that we are well past 1.4?

retep998:
First I stop procrastinating and land the various PRs that are hanging around in limbo. Then once I'm caught up I experiment with removing `::` everywhere.

retep998:
Actually I think I'd like to move things further into submodules based on the subfolder names so it would be `winapi::um::d3d11` for example instead of just `winapi::d3d11`. This would break any code that refers to the submodules explicitly so I'll have to wait for winapi 0.3 for this change.

retep998:
Replaced by https://github.com/retep998/winapi-rs/issues/316

PR #100, opened 2015-05-21
COM Helpers and Various Constants

excaliburHisSheath:
- Adds the RIDL declarations for IMMDevice, IMMDeviceEnumerator, IAudioClient, and IAudioRenderClient.
- Adds `SetPriorityClass()` to kernel32-sys.
- Adds `timeBeginPeriod()` and `timeEndPeriod()` to winmm-sys.
- Adds various constants related to core audio, timers, and thread priorities.

I left the RIDL declarations in the main winapi lib file, but if you want them each in a separate file like you did with unknownbase.rs then I can move them over.


retep998:
As issue #17 states, everything is being split out of `lib.rs` into separate files for each header.

src/lib.rs
```diff
@@ -1234,87 +1290,69 @@ pub const IID_IMMDeviceEnumerator: IID = GUID {
     Data4: [0xA7, 0x46, 0xDE, 0x8D, 0xB6, 0x36, 0x17, 0xE6],
 };
 
-#[repr(C)]
-pub struct IMMDeviceVtbl {
-    pub QueryInterface: unsafe extern "system" fn(
-        This: *mut IMMDevice,
+RIDL!(
+interface IMMDevice(IMMDeviceVtbl) {
```
retep998: Could you change the COM definitions to also inherit from their base COM classes, like the way I did with `AsyncIUnknown`?

excaliburHisSheath: No problem, it's taken care of.

excaliburHisSheath:
I moved all the things I added or changed into the appropriate modules.

src/audioclient.rs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License <LICENSE.md>
 //! this ALWAYS GENERATED file contains the definitions for the interfaces
 //1627
+use unknwnbase::*;
```
retep998: I'd prefer not to do this.

excaliburHisSheath: If I try to do `::IUnknown` the RIDL macro doesn't compile, so I need to import the type manually. Would it be better to change the macro so that it supports absolute paths?

retep998: It would be better to fix the macro.

excaliburHisSheath: This should be fixed now.

retep998:
Thanks, could you rebase everything and maybe squash it a bit?

excaliburHisSheath:
Rebased and squashed. Let me know if it looks like I did anything wrong, I don't normally rebase/squash in my projects so I wasn't familiar with the process.

PR #107, opened 2015-06-11
Add wave* and some time* functions

jminer:
Is this a good way to do unions (`MMTIME`)? It is similar to how bindgen translated the union.

### DIFF
--- a/src/mmsystem.rs
+++ b/src/mmsystem.rs
@@ -1,8 +1,212 @@
 // Copyright © 2015, Peter Atashian
 // Licensed under the MIT License <LICENSE.md>
+//! MM procedure declarations, constant definitions and macros
+//109 (Win 7 SDK)
+pub type MMVERSION = ::UINT;
 pub type MMRESULT = ::UINT;
-
+#[repr(C)] #[derive(Clone, Copy, Debug)]
+pub struct MMTIME {
+    wType: ::UINT,
+    u: MMTIME_u,
+}
+pub type PMMTIME = *mut MMTIME;
+pub type NPMMTIME = *mut MMTIME;
+pub type LPMMTIME = *mut MMTIME;
+#[repr(C)] #[derive(Clone, Copy, Debug)]
+pub struct MMTIME_u {
+    data: [u8; 8],
+}
+UNION!(MMTIME_u, data, ms, ms_mut, ::DWORD);
+UNION!(MMTIME_u, data, sample, sample_mut, ::DWORD);
+UNION!(MMTIME_u, data, cb, cb_mut, ::DWORD);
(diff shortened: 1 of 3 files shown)


src/mmsystem.rs
```diff
@@ -1,8 +1,230 @@
...
+pub const CALLBACK_EVENT: ::DWORD = 0x00050000;
+//497 (Win 7 SDK)
+pub const WAVERR_BADFORMAT: MMRESULT = WAVERR_BASE + 0;
+pub const WAVERR_STILLPLAYING: MMRESULT = WAVERR_BASE + 1;
+pub const WAVERR_UNPREPARED: MMRESULT = WAVERR_BASE + 2;
+pub const WAVERR_SYNC: MMRESULT = WAVERR_BASE + 3;
+pub const WAVERR_LASTERROR: MMRESULT = WAVERR_BASE + 3;
+#[repr(C)] #[derive(Clone, Copy, Debug)]
+pub struct HWAVEIN__;
+pub type HWAVEIN = *mut HWAVEIN__;
```
retep998: Use `DECLARE_HANDLE!(HWAVEIN, HWAVEIN__);`

retep998:
I'm not sure how I want to deal with untagged unions. Ideally Rust would get something like https://github.com/rust-lang/rfcs/pull/724 which would makes things way easier. I'm not a fan of those methods returning a `*mut` instead of a `&mut`, plus if all you have is an immutable `&` then you can't access those union members at all.

src/mmsystem.rs
jminer: Fixed `HWAVEIN` and `HWAVEOUT`.

jminer:
Yeah, I wish there were language support for C unions. What do you think about having `mut` and non-`mut` versions of each member?

``` rust
    pub unsafe fn ms<'a>(&'a self) -> &'a ::DWORD {
        ::std::mem::transmute(&self.data)
    }
    pub unsafe fn ms_mut<'a>(&'a mut self) -> &'a mut ::DWORD {
        ::std::mem::transmute(&mut self.data)
    }
```

The methods should probably have an `inline` attribute too.

retep998:
There shouldn't be any need to explicitly specify the lifetime in this case as lifetime inference should be enough.

jminer:
Right, thanks. I was thinking that `transmute` doesn't preserve the lifetime, but I forgot that's irrelevant as the body of a method doesn't affect lifetime elision. I updated the commit.

retep998:
Do you think you could update this PR to try to make use of the `UNION!` macro I wrote?

jminer:
Sure, it's updated. I think using a macro is a good improvement.

PR #109, opened 2015-06-11
Fill out some dbghelp bindings

alexcrichton:
Unfortunately MinGW can't actually link to `dbghelp` so `dbghelp-sys` won't be too useful yet, but the type definitions are nice to have!


alexcrichton:
I added a few more assorted functions, and I think this is good to go.

alexcrichton:
Thanks @retep998!

PR #112, opened 2015-06-14
Add usp10

jminer:
I plan on adding the other Uniscribe functions and structures, but first I wanted to get feedback on how to do bit fields, as Uniscribe uses a lot of them. Two unsupported C features in one week...


retep998:
Oh, oh my.

jminer:
I added the rest of Uniscribe and updated the `BIT_FIELDS` macro a little. I would have had one bit field per macro call, like the `UNION` macro, but then the lines are over 100 chars.

Now the macro takes the type of the bit field and casts to and from it. I thought of having one of the bit fields be `SCRIPT_JUSTIFY` and the single-bit bit fields be `bool`; however, Rust doesn't allow casting to an `enum` or `bool`. I translated `SCRIPT_JUSTIFY` values as `pub const`s for now so that they don't require casts. Would you prefer having `enum` and `bool` bit field types?

retep998:
I'm gonna experiment with the BIT_FIELDS macro a bit until I come up with something that looks nice.

retep998:
I have made a `BITFIELD` macro which I think looks nice.

jminer:
Great, thanks! I updated the PR. I think I prefer passing the length of the bit field to the macro instead of the end so that it is easier to compare to the header file, but I do like how the range looks.

PR #114, opened 2015-06-20
Add basic vssapi support

bvinc83:
Not all of the VSS api is finished, but at its current state, it is usable.

This lets you link statically against the microsoft vssapi.lib import library.  But this library is a bit different.  All symbols in the import library are c++ mangled names and need to be specified with the link_name attribute.  On x86_64, I would always hit rust issue 23216 and it would crash.  On x86, I had to discover the trick of starting the link_name attribute with \x01, or else rust would further mangle the specified link_name attribute and it would fail to link.

One struct used a field named "type" that I had to rename "type_".

I hit one c union that will probably need to be cleaned up later.

I ran into a lot of c++ enum types.  Initially I modeled them with rust enums, but now I don't think rust enums are a good fit.  They're more cumbersome to use because they have their own namespace, and they don't let you apply the | operator to them, and about half of them are flags that need the | operator.  C++ doesn't guarantee a size of a normal enum, but from what I can tell, they're always i32 in Microsoft's ABI, right?  I settled to using INT or DWORD depending on if they were numbers or flags.  This needs more thought.  Maybe they should all be the same type, but sometimes they're negative and sometimes they're >= 2^31.


retep998:
For flags using `DWORD` is fine, but for enums that aren't flags, please use rust enums. You can do `pub use SOME_ENUM::*;` to replicate C enum contents not being namespaced. Enums should use `#[repr(i32)]`. Values being outside of the valid range of a type can be handled with a simple cast `Blah: DWORD = -1i32 as u32;`.

Also, please try to match the existing style by not vertically aligning things.

src/vsbackup.rs
```diff
@@ -0,0 +1,219 @@
...
+    cbIcon: ::UINT,
+    bRestoreMetadata: bool,
+    bNotifyOnBackupComplete: bool,
+    bSelectable: bool,
+    bSelectableForRestore: bool,
+    dwComponentFlags: ::DWORD,
+    cFileCount: ::UINT,
+    cDatabases: ::UINT,
+    cLogFiles: ::UINT,
+    cDependencies: ::UINT,
```
retep998: Missing `pub` on these fields.

src/vss.rs
```diff
@@ -0,0 +1,248 @@
...
+pub const VSS_VOLSNAP_ATTR_TXF_RECOVERY: ::DWORD         = 0x02000000;
+pub const VSS_VOLSNAP_ATTR_FILE_SHARE: ::DWORD           = 0x04000000;
+
+pub type VSS_SNAPSHOT_CONTEXT = ::DWORD;
+pub type PVSS_SNAPSHOT_CONTEXT = *mut VSS_SNAPSHOT_CONTEXT;
+pub const VSS_CTX_BACKUP: ::DWORD                    = 0;
+pub const VSS_CTX_FILE_SHARE_BACKUP: ::DWORD         = VSS_VOLSNAP_ATTR_NO_WRITERS;
+pub const VSS_CTX_NAS_ROLLBACK: ::DWORD              = ( ( VSS_VOLSNAP_ATTR_PERSISTENT | VSS_VOLSNAP_ATTR_NO_AUTO_RELEASE ) | VSS_VOLSNAP_ATTR_NO_WRITERS );
+pub const VSS_CTX_APP_ROLLBACK: ::DWORD              = ( VSS_VOLSNAP_ATTR_PERSISTENT | VSS_VOLSNAP_ATTR_NO_AUTO_RELEASE );
+pub const VSS_CTX_CLIENT_ACCESSIBLE: ::DWORD         = ( ( ( VSS_VOLSNAP_ATTR_PERSISTENT | VSS_VOLSNAP_ATTR_CLIENT_ACCESSIBLE ) | VSS_VOLSNAP_ATTR_NO_AUTO_RELEASE ) | VSS_VOLSNAP_ATTR_NO_WRITERS );
```
retep998: No need for all these parentheses. Also use line breaks to fit within 99 columns.

src/vss.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub m_pwszProviderVersion: ::VSS_PWSZ,
+    pub m_ProviderVersionId: ::VSS_ID,
+    pub m_ClassId: ::CLSID,
+}
+type PVSS_PROVIDER_PROP = *mut VSS_PROVIDER_PROP;
+
+#[repr(C)] #[derive(Clone, Copy, Debug)]
+pub struct VSS_OBJECT_PROP {
+    pub Type: VSS_OBJECT_TYPE,
+    pub union_: VSS_SNAPSHOT_PROP, // This is a union of VSS_SNAPSHOT_PROP and VSS_PROVIDER_PROP
```
retep998: You can use the `UNION!` macro to do union stuff.

bvinc83:
Thanks for reviewing it.  I fixed everything.

The Travis CI build failure isn't my fault.  It's a missing close parenthesis on lib/secur32-sys/src/lib.rs:114

retep998:
Ha, whoops, that build failure was my fault. :stuck_out_tongue:

PR #118, opened 2015-06-24
add time types and functions

seanmonstar:


seanmonstar:
Um, wut? `FILETIME` and `SYSTEMTIME` are already defined elsewhere? They are in `winbase.h`, and I didn't find definitions in `winbase.rs`....

seanmonstar:
Ok, so `FILETIME` is in `minwindef.rs`, and `SYSTEMTIME` is in `minwinbase.rs`. Should they stay there, or be moved to `winbase.rs`?

PR #130, opened 2015-07-10
Add WM_NCHITTEST and MOUSEHOOKSTRUCT codes.

diaphore:
Converted from `<WinUser.h>`
The constants are defined as `c_int` because that's what they are in C but `MOUSEHOOKSTRUCT::wHitTestCode` is a `UINT` and `WM_NCHITTEST` returns those through a `LRESULT`.


retep998:
I think I'd prefer if these used `UINT` since they're going to be used most commonly in `MOUSEHOOKSTRUCT` anyway.

PR #136, opened 2015-07-15
Add a new macro for multiple inheritance.

bvinc83:
How do you feel about 2 new macro patterns for RIDL:

1.  Support multiple inheritance.
2.  Support multiple inheritance with no new methods.

It's a little messier since you have to specify the names of the fields.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,6 +188,44 @@ macro_rules! RIDL {
             }
         }
     };
+    (interface $interface:ident ($vtbl:ident)
+        : $( $pinterface:ident ($pvtbl:ident, $pfield:ident) ),*
+        {$(
+            fn $method:ident(&mut self $(,$p:ident : $t:ty)*) -> $rtr:ty
+        ),+}
+    ) => {
+        #[repr(C)] #[allow(missing_copy_implementations)]
+        pub struct $vtbl {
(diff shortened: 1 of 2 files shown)


retep998:
According to MSDN...

>The IVssWriterComponentsExt interface is a C++ (not COM) interface used by requesters to access and modify the components of a writer involved in a backup.

Can you find any examples of multiple inheritance in actual COM interfaces?

bvinc83:
No, I can't.  And looking at MIDL, it looks like it does not support multiple inheritance.

The pattern that I keep seeing is that a VSS interface gets an Ex tacked on to the end, just to support the IUnknown interface (using multiple inheritance).

Oh god, I assumed that all of these interfaces that inherited from IUnknown were COM interfaces, but it  looks like VSS is a mix of COM interfaces and C++ interfaces.  But all of the C++ interfaces appear to be binary compatible with COM (except maybe these multiple inheritance ones).  Using the RIDL macro has been super convenient so far.  But the RIDL macro was only meant for COM interfaces, right?

Hmm.  I'm not sure how to proceed.  Should I undo the RIDL macro from the non-COM objects?  Maybe there should be a macro to help define c++ pure virtual classes, and the RIDL macro can use it?

PR #138, opened 2015-07-15
Kernel32: Add some more functions

CruzBishop:
GetTickCount(+64) are required for a project still in development, and the others will be needed in about a month or so if I get it off the ground.

I'll have some time off soon, so I'll try to aim for larger PRs.


CruzBishop:
Updated with some randomly selected functions

PR #144, opened 2015-07-16
Copyfile

### LINKED ISSUE #126: Why are the VK_KEY constants of type WPARAM
These are single byte values.

Plus WPARAM is a different size depending on Win 32-bit vs 64-bit which could easily break people's code if they don't have both 32 and 64 bit to test on.

### LINKED ISSUE #139: mmap constants are missing
The Following mmap constants are missing, as documented [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366761(v=vs.85).aspx):

* `FILE_MAP_READ`
* `FILE_MAP_WRITE`
* `FILE_MAP_ALL_ACCESS`
* `FILE_MAP_COPY`
* `FILE_MAP_EXECUTE`

iliekturtles:
Number of small updates. Let me know if you would like the PR squashed or any of the commits excluded.

Closes #139
Closes #126 (Corrects VK_* parameters to be c_int, https://msdn.microsoft.com/en-us/library/windows/desktop/ms646301(v=vs.85).aspx)


lib/gdi32-sys/src/lib.rs
```diff
@@ -356,7 +356,7 @@ extern "system" {
         hdc: HDC, hbm: HBITMAP, start: UINT, cLines: UINT, lpvBits: LPVOID, lpbmi: LPBITMAPINFO,
         usage: UINT
     ) -> c_int;
-    // pub fn GetDeviceCaps();
+    pub fn GetDeviceCaps(hdc: HDC, nIndex: i32) -> i32;
```
retep998: `c_int`, not `i32`

src/wingdi.rs
```diff
@@ -195,7 +195,48 @@ pub type LPLOGFONTW = *mut LOGFONTW;
 pub fn RGB (r: ::BYTE, g: ::BYTE, b: ::BYTE) -> ::COLORREF {
   r as ::COLORREF | ((g as ::COLORREF) << 8) | ((b as ::COLORREF) << 16)
 }
-//1906 (Win 7 SDK)
+//
+pub const DRIVERVERSION: i32 = 0;
```
retep998: `c_int` for these as well

iliekturtles:
`i32` references fixed.

diaphore:
Regarding `VK_*`, they can also be passed as :
* `UINT` https://msdn.microsoft.com/en-us/library/windows/desktop/ms646306%28v=vs.85%29.aspx
* `DWORD` https://msdn.microsoft.com/en-us/library/windows/desktop/ms644967%28v=vs.85%29.aspx

Let's just stick to `u8` promoted to `c_int`.

retep998:
Oh how I wish Rust had a way to represent integer consts that will implicitly coerce to larger integer types.

PR #149, opened 2015-07-17
OpenGL: Round one of additions

CruzBishop:
This PR does the following:

 * Adds some constants and most types from gl.h to gl.rs
 * Enables glAccum

I'm too tired to do any more right now, so I'll come back in a few days :)


lib/opengl32-sys/src/lib.rs
```diff
@@ -11,8 +11,8 @@ extern "system" {
     // pub fn GlmfEndPlayback();
     // pub fn GlmfInitPlayback();
     // pub fn GlmfPlayGlsRecord();
-    // pub fn glAccum();
-    // pub fn glAlphaFunc();
+    pub fn glAccum(op: GLenum, value: GLfloat);
+    // pub fn glAlphaFunc(func: GLenum, ref: GLclampf);
```
CruzBishop: I ran into trouble here, since ref is a reserved keyword. Is there a way to get around this and, if not, can I change this to _ref or reference and still have the FFI binding work?

retep998: The parameter name has no effect on the functionality of the FFI binding. Feel free to rename it to something that is similar but isn't a keyword.

tomaka:
Note that the `gl-rs` library is more suitable for OpenGL, for two reasons:
 - OpenGL functions are the same whatever the platform is (windows, linux, android, etc.)
 - gl-rs handles all OpenGL versions and extensions (including WGL), whereas here you'd just have OpenGL 1.1

CruzBishop:
@retep998 I updated glAlphaFunc, sorry about the delay.

@tomaka Thanks! I'll take a look at it, I was just working on this since the crate already had some parts completed

PR #157, opened 2015-07-22
Add [Get,Set][Window,Class]LongPtr[A,W] for x86

diaphore:
The `*LongPtr*` functions are #defined to `*Long*` for non-`_WIN64` builds but defining them within `winapi` would introduce a dependency to `user32-sys` so I put them into the import lib which is non-canonical ... Nasty one way or the other.

### DIFF
--- a/lib/user32-sys/src/lib.rs
+++ b/lib/user32-sys/src/lib.rs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License <LICENSE.md>
 //! FFI bindings to user32.
 #![cfg(windows)]
+#![allow(non_snake_case)]
 extern crate winapi;
 use winapi::*;
 extern "system" {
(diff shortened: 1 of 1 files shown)


retep998:
At the moment I'm avoiding aliases for functions, although I may think about it sometime later down the road.

diaphore:
They're part of the API though. Would you prefer fitting all the aliases in a separate crate ?

retep998:
There's also the matter of the aliases based on `UNICODE`, aliases based on which version of Windows you're targeting. Quite a few unanswered questions about how they should be handled, so I'm putting it off for now. Since without aliases you can still use all the functions, although you might need to use `cfg` a bit, it's not a pressing issue.

diaphore:
Sure, no hurry. I was just wondering if you had any plan.

PR #160, opened 2015-07-23
Implement std::fmt::Display for GUID

gentoo90:
```rust
print!("{}", guid);
```
will print the string representation of `GUID`


retep998:
At the moment the only traits that I'm implementing are `Copy` `Clone` and `Debug`. It should be easy enough to implement `Display` in your own crate for a newtype around `GUID`.

retep998:
Closing as out of scope for now.

PR #162, opened 2015-07-26
More Vss work.  vsbackup is complete

bvinc83:
Fix some types LPWSTR -> LPWCSTR
Remove an instance of usize that should never be in FFI code


lib/vssapi-sys/src/lib.rs
```diff
@@ -2,19 +2,53 @@
...
+#[inline]
+pub unsafe fn GetProviderMgmtInterface(
+    ProviderId: VSS_ID, InterfaceId: IID, ppItf: *mut *mut IUnknown
+) -> HRESULT {
+    GetProviderMgmtInterfaceInternal(ProviderId, InterfaceId, ppItf)
+}
+#[inline]
+pub unsafe fn ShouldBlockRevert(wszVolumeName: LPCWSTR, pbBlock: *mut bool) -> HRESULT {
+    ShouldBlockRevertInternal(wszVolumeName, pbBlock)
+}
```
retep998: Could you leave out these inline functions for now? I want the `-sys` crates to just be extern declarations for now.

retep998:
Could you rebase this PR against master?

lib/vssapi-sys/src/lib.rs
bvinc83: These are super important inline functions.  They're the entry point into the rest of the API.

In the SDK, they are inline functions that exist inside of vsbackup.h to wrap around the actual symbol names *Internal, that you aren't supposed to call directly.

Would you rather they exist in src/vsbackup.rs?

Or I could deviate from the SDK and use the link_name attribute instead of inline wrapper functions.

retep998: I think I'd rather have `#[link_name]` be used here.

bvinc83:
Fixed to use the link_name attribute, and rebased

PR #163, opened 2015-07-26
Fixed handle type to foreign-function safe

skdltmxn:
It looks like new rust compiler is giving warnings about the type DECLARE_HANDLE generates.
I changed *non-foreign-function safe* type to foreign-function safe.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,7 +107,7 @@ pub use xinput::*;
 //-------------------------------------------------------------------------------------------------
 macro_rules! DECLARE_HANDLE {
     ($name:ident, $inner:ident) => {
-        #[repr(C)] #[allow(missing_copy_implementations)] struct $inner { unused: () }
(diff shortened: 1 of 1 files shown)


retep998:
Could we perhaps use c_void instead?

skdltmxn:
I used c_int because MS SDK implemented HANDLE with int.
Do you think c_void would be better?

retep998:
It would help to hammer in the concept that a `HANDLE`-like thing should never be constructable by user code, since `c_void` cannot be instantiated.

skdltmxn:
Oh, that's fair point. Changed to c_void!

diaphore:
[A discussion](https://github.com/rust-lang/rust/issues/27303) and [an example](https://github.com/rust-lang/cargo/pull/1836/files). They seem to recommend `enum Whatever {}` for some reason.

PR #166, opened 2015-07-30
Add comctl32 support

skdltmxn:
I added support for comctl32


src/prsht.rs
```diff
@@ -0,0 +1,309 @@
...
+// Licensed under the MIT License <LICENSE.md>
+//! Interface for the Windows Property Sheet Pages
+#[repr(C)] #[allow(missing_copy_implementations)]
+pub struct PSP {
+    unused: ::c_void,
+}
+pub type HPROPSHEETPAGE = *mut PSP;
+
+#[repr(C)] #[allow(missing_copy_implementations)]
+pub struct _PROPSHEETPAGEA {
```
retep998: Could you instead have this be `PROPSHEETPAGEA` and just have it be a typedef to `PROPSHEETPAGEA_V4`?

retep998:
In https://github.com/skdltmxn/winapi-rs/blob/comctl32/src/commctrl.rs#L149 you have extra parentheses that aren't needed. I'd comment on the diff, but the diff was too big for github to display it. :disappointed:

skdltmxn:
I removed parentheses and _PROPSHEETPAGEA since there is typedef below.

retep998:
Could you rebase this PR, and your other PRs, against master?

skdltmxn:
Ok, it's rebased

retep998:
Your msvc tests appear to be testing functions that don't exist at runtime on appveyor. Could you figure out which symbols they are and disable them from testing?

skdltmxn:
hmm it's strange, I can't find any failure from msvc test logs but only a message saying "Build execution time has reached the maximum allowed time for your plan (40 minutes)."

retep998:
The issue is that if a program links to a symbol which doesn't actually exist in the DLL, then at runtime the process will error. How the error manifests depends on the environment it is run in, and apparently in the appveyor environment it pops up a dialog saying there is a missing symbol. Since it is automated CI and there is nobody to click that dialog it just hangs until it times out.

skdltmxn:
Resolved!

PR #167, opened 2015-07-31
build: avoid panicing build script if target does split to >2 elements

jmesmon:
In meta-rust, I've observed failures caused by missing the first length check.

I haven't seen any failures caused by the second (because I'm not building anything for windows) but I don't see a reason not to check for it.


retep998:
Thanks for catching this.

PR #170, opened 2015-08-01
Start oleaut32 with a few functions

bvinc83:


retep998:
Why are a bunch of the functions commented out even though they have full signatures?

CruzBishop:
Perhaps not 100% certain?

bvinc83:
I did all of the work for their signature, but not the work to implement the crazy complicated types like VARIANT.  The really important ones to have are the first ones like SysAllocString.

lib/oleaut32-sys/src/lib.rs
```diff
@@ -1,9 +1,382 @@
 // Copyright © 2015, Peter Atashian
 // Licensed under the MIT License <LICENSE.md>
-//! FFI bindings to oleaut32.
-#![no_std]
-#![experimental]
+//! FFI bindings to vssapi.
```
retep998: Are you sure you intended to change this comment?

bvinc83:
Good catch on the comment.  Fixed and rebased to master.

PR #175, opened 2015-08-05
Add psapi functions

skdltmxn:
I added remaining psapi functions


skdltmxn:
Rebased against master

PR #176, opened 2015-08-08
Add WinMM MIDI support

Boddlnagg:
This adds all functions (and necessary types) related to MIDI support in WinMM.
I had to add LPUINT to the basic definitions ... don't know if this was missing for a reason?


retep998:
`LPUINT` isn't defined in `minwindef`. It is actually in either `imm` or `immdev`, at least according to the Windows 10 SDK.

Boddlnagg:
Okay, so I created `src/imm.rs` and moved the definition there. Is this correct now?

PR #179, opened 2015-08-13
Add to winioctl.rs

bvinc83:


src/winioctl.rs
```diff
@@ -88,6 +88,124 @@ pub const FILE_ANY_ACCESS: ::DWORD = 0;
 pub const FILE_SPECIAL_ACCESS: ::DWORD = FILE_ANY_ACCESS;
 pub const FILE_READ_ACCESS: ::DWORD = 0x0001;
 pub const FILE_WRITE_ACCESS: ::DWORD = 0x0002;
+//281
+pub const IOCTL_STORAGE_BASE: ::DWORD = FILE_DEVICE_MASS_STORAGE;
+pub const IOCTL_STORAGE_CHECK_VERIFY: ::DWORD = CTL_CODE!(IOCTL_STORAGE_BASE, 0x0200, METHOD_BUFFERED, FILE_READ_ACCESS);
```
retep998: Maybe we could use newlines so these can fit within 99 columns?

bvinc83: Yeah.  I thought maybe this file was an exception to the rule.  I'll change it.

bvinc83:
I added line breaks and rebased to master

PR #185, opened 2015-08-19
Better appveyor configuration

retep998:
Test all four windows target triples for maximum coverage


retep998:
This is a genuinely worrying issue that the kernel32 test is failing _only_ on 32-bit msvc.

retep998:
I traced the issue down to the function `RtlPcToFileHeader`. For some reason it isn't available on some versions of Windows in the 32-bit DLLs. It is on Windows 10 though, which made finding the issue locally difficult.

Issue #196, opened 2015-09-01
Need to handle alignment


retep998:
The Windows SDK headers define alignment and packing for a lot of things, and so far `winapi` hasn't really dealt with it. There's likely quite a few incorrect definitions hanging around that need to be fixed.

PR #202, opened 2015-09-07
Use #[repr(packed)] for WAVEFORMATEX and WAVEFORMATEXTENSIBLE

tomaka:
There's this strange GUID alignment problem without this fix.

cc #196 

I know that you are planning to check everything, but modifying these two ahead of time can't hurt.


retep998:
It has to be `#[repr(C, packed)]`.

tomaka:
Updated

PR #204, opened 2015-09-11
Add netapi32 support

skdltmxn:


retep998:
Do you think you could use the new `ENUM!` and `FLAGS!` macros for your enums?

skdltmxn:
It looks like ENUM! and FLAGS! both require every entry of the enum to have specific value.
How can I use those macros for enums without any values?

retep998:
The value of each enum variant is just the previous + 1, and the first, if not specified, is 0. Making the macro handle that automatically would involve macro recursion and that's a bit beyond my current abilities with macros.

retep998:
Actually, with a bit of help, I managed to improve the `ENUM!` macro to be able to calculate implicit values for the variants. `FLAGS!` still needs explicit values, but that's okay since the variants are powers of two, not sequential integers, and C doesn't implicitly do that either.

skdltmxn:
I replaced my enums with ENUM! macro.

PR #210, opened 2015-09-14
DX12, DXGI1_2, DXGI1_3, DXGI1_4

red75prime:
Featuring some bugfixes


msiglreith:
Oh nice, this looks more complete than mine.

src/d3d12.rs
```diff
@@ -0,0 +1,2285 @@
...
+        ppvCommandSignature: *mut *mut ::c_void
+    ) -> ::HRESULT,
+    fn GetResourceTiling(
+        &mut self, pTiledResource: *mut ::ID3D12Resource, pNumTilesForEntireResource: *mut ::UINT,
+        pPackedMipDesc: *mut ::D3D12_PACKED_MIP_INFO,
+        pStandardTileShapeForNonPackedMips: *mut ::D3D12_TILE_SHAPE,
+        pNumSubresourceTilings: *mut ::UINT, FirstSubresourceTilingToGet: ::UINT,
+        pSubresourceTilingsForNonPackedMips: *mut ::D3D12_SUBRESOURCE_TILING
+    ) -> (),
+    fn GetAdapterLuid(&mut self, __ret_val: *mut ::LUID) -> *mut ::LUID
```
msiglreith: This signature looks wrong (https://msdn.microsoft.com/en-us/library/windows/desktop/dn914411%28v=vs.85%29.aspx)

retep998: Nope, it's actually correct since `stdcall` actually differs depending on whether it is a method or a function. In particular in certain cases when returning an aggregate, it actually takes an extra parameter which is a pointer to where to store the return value, and it also returns that pointer. This was carefully tested.

msiglreith: Interesting, thanks for clearing it up!

red75prime:
Error: unions are not processed.

retep998:
I manually squashed and landed this as https://github.com/retep998/winapi-rs/commit/0ba80f1b11f7833ab8deaa785ea6b07a97b48723

PR #217, opened 2015-10-06
Clean whitespace

aarzee:
Remove leading newlines; replace lines containing only whitespace with empty lines; replace multiple trailing newlines with a single newline


retep998:
I think you got this PR a bit wrong. You ended up replacing blank lines which use \r\n for a newline with a \n while all the other lines still use \r\n.

aarzee:
My script doesn't account for \r\n... I'll update it and try again later.

PR #220, opened 2015-10-08
Clean whitespace

aarzee:


retep998:
Unfortunately the one change that this PR makes is no longer relevant. Thanks for the effort though!

PR #221, opened 2015-10-12
Add #![no_std] Support

excaliburHisSheath:
In theory winapi-rs should be able to work with `#![no_std]`, however there are a few cases where `std` is used that prevents winapi-rs from being used with a `#![no_std]` crate:

- `std::raw::os` is used to get C types without depending directly on `libc`. To get around this we simply copy the definitions from `std::raw::os`. This is a potential source of issues if the definitions in `std` are updated and the local ones aren't, but that seems unlikely since C types aren't likely to change any time soon.
- `std::mem` and `std::ops` are used within the `RIDL!{}` macro to make using COM types more ergonomic. `mem` and `ops` are already being re-exported by `std` from `core`, however we can't just use `core` directly in winapi because `core` is unstable and can only be used on nightly. To get around this we create a local `std` module to re-export those features and only compile it when `#![no_std]` is used.

To allow users to choose whether they want `#![no_std]` support I've added a `"no_std"` feature.

I don't know if this feature is desirable, so I've only done a proof of concept in this initial PR. If it is decided that this feature is worth adding there are a few more things that will need to be done before this is ready to merge:

- The `"no_std"` feature will need to be added to all of the -sys crates, as well as the conditional compilation flags for applying `#![no_std]`.
- Additional settings will need to be added to the continuous integration settings for Travis and Appveyor so that this `"no_std"` feature can be tested.

### Unresolved Questions

- Should the dummy `std` module be left inline in `lib.rs` or should it be moved to a separate file.
- Will `RIDL!{}` and the other macros that rely on `std::mem`/`std::ops` ever be used outside of `winapi`? If so any references to `std` will need to be replaced with `$crate::std`.


retep998:
Theoretically yes, I did actually have winapi using `no_std` before, however it isn't entirely practical. Since this library will only ever be used on Windows, which isn't really an embedded platform, there isn't really a need to not use std, nor are there any alternate std libraries. As such I'd rather wait until if and when `no_std` becomes stable.

excaliburHisSheath:
Fair enough. I was only playing with `no_std` on my machine as an experiment anyway, so it's not like I have a valid use case.

PR #224, opened 2015-10-19
Clean whitespace

aarzee:
Remove leading newlines; replace lines containing only whitespace with empty lines; replace multiple trailing newlines with a single newline; remove trailing whitespace in lines.

This PR was created semiautomatically.


retep998:
`libsecur32.a` is a binary file and should not have been modified

aarzee:
Thanks, but if that's the case, why is it in version control? I'll make a new PR.

retep998:
It's in version control because I have to bundle import libraries with some of the -sys crates when used with -gnu targets because I cannot assume the presence of a full MinGW installation.

PR #228, opened 2015-10-25
use STRUCT! ware it is simple

Eh2406:
used regexp to fined cases ware STRUCT! is simple

### DIFF
--- a/src/cfgmgr32.rs
+++ b/src/cfgmgr32.rs
@@ -99,57 +99,51 @@ pub const fMD_NonCacheable: ::DWORD = 0x0;
 pub const fMD_Cacheable: ::DWORD = 0x20;
 pub const fMD_WINDOW_DECODE: ::DWORD = 0x40;
 pub const fMD_MEMORY_BAR: ::DWORD = 0x80;
-#[repr(C)] #[derive(Clone, Copy, Debug)]
-pub struct MEM_RANGE {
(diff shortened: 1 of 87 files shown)


retep998:
ware -> where
fined -> find

Eh2406:
Sorry about the miss spellings.

Is there a category you want me to look at next? struct with #[derive(Copy)] and impl Clone for ... { fn clone(&self) -> ... { *self } }? or enums?

retep998:
Any Rust enums need to be switched over to `ENUM!` or `FLAGS!` as appropriate. That's all I can think of at the moment. I'll deal with the remaining structs myself once I decide how to deal with them.

Eh2406:
Is there an idiot proof way to determine which one to use? If not I can start looking them up one at a time and see how far I get.

retep998:
If the discriminants are powers of two, then it is `FLAGS!`. Otherwise it is likely `ENUM!`.

Eh2406:
most rust `enom`'s are `#[repr(i32)]` the macro uses `u32` is it ok for me to replace? (sorry to be constantly bugging you. I just want to get it right, so as to not make more work fixing my mistakes.)

retep998:
The representation that the macros uses is fine, so feel free to replace all enums with them regardless of what the repr is.

PR #240, opened 2015-11-10
DXGI&DWrite IIDs and funcs' signatures fixes

red75prime:


retep998:
I've actually filled out `dxguid-sys` completely recently so could you remove the redundant IID declarations here?

PR #244, opened 2015-11-25
fix DECLARE_HANDLE

th0razine:
This fixes structs generated by the DECLARE_HANDLE macro being private.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -3,7 +3,7 @@
 //! Macros to make things easier to define
 macro_rules! DECLARE_HANDLE {
     ($name:ident, $inner:ident) => {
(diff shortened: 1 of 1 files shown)


retep998:
There shouldn't be any reason to have to refer to the inner struct. Any usage of the inner type, such as `HWND__` is always done through a `*mut` which has a public alias such as `HWND`. Is there any reason why you can't just use `HWND` instead of trying to refer to `HWND__`?

th0razine:
Yea, it would seem that it was user error on my end. `winapi::HWND` is definitely callable - rustc likes to refer to it as `winapi::windef::HWND__`, however -- hence the confusion.

PR #248, opened 2015-12-12
Fix an interface signature in d2d1.rs

Connorcpu:
Looks like I made a mistake here, because the real signature is void https://msdn.microsoft.com/en-us/library/windows/desktop/dd316911%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396


retep998:
I hope you don't mind that I rebased your commits for you.

PR #259, opened 2016-02-21
Added clipboard constants to winuser.rs

Osspial:
From here: https://msdn.microsoft.com/en-us/library/windows/desktop/ff729168%28v=vs.85%29.aspx


retep998:
Could you squash these two commits together?

Osspial:
Done.

PR #260, opened 2016-02-23
D3D10 shader module

sectopod:


sectopod:
I don't think it makes sense to merge this until we figure out https://github.com/gfx-rs/gfx/pull/861
At the end of the day, all DX10 class hardware supports DX11 interfaces anyway (just at a lower feature level).

retep998:
It makes sense to have this regardless as this is more bindings for more APIs, which is always good.

retep998:
Appveyor failing due to https://github.com/rust-lang/rust/issues/31840

Issue #274, opened 2016-04-01
d3d11.rs - wrong signature for ID3D11DeviceContext::ExecuteCommandList method


pravic:
```rust
fn ExecuteCommandList(&mut self, pCommandList: *mut ID3D11CommandList) -> ()
```
vs
```c
        void ( STDMETHODCALLTYPE *ExecuteCommandList )( 
            ID3D11DeviceContext2 * This,
            /* [annotation] */ 
            _In_  ID3D11CommandList *pCommandList,
            BOOL RestoreContextState);
 ```
[msdn](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476423(v=vs.85).aspx)

It was caught [here](https://github.com/gfx-rs/gfx/issues/923) with nightly-i686-msvc. IDK about another places with similar problems.

pravic:
bump. Or PR is required for such issues?

retep998:
For something small like this a PR isn't necessary. It's just that I haven't gotten around to dealing with it yet.

pravic:
Ping?

kvark:
@pravic if you are doing GFX work and blocked by this, you can easily work around the problem by switching to emulated command buffers (instead of native ones)

retep998:
Or using a local cargo override.

pravic:
This was fixed in 0.2.7, which crates.io has, but github master is 0.2.6. I am in doubt.

retep998:
@pravic It was on the dev branch (which I just pushed to master). I have it configured so I can't push to master unless it first passes CI on dev, which helps to prevent accidentally breaking master.

PR #277, opened 2016-04-15
Update d3d11.rs ExecuteCommandList

Eh2406:
fix for #274


Eh2406:
updated to fix extra ","

PR #278, opened 2016-04-17
Revert STRUCT! definition in mmreg

mvdnes:
`WAVEFORMATEX` and `WAVEFORMATEXTENSIBLE` need `#[repr(C, packed)]`, see #202.


mvdnes:
Any update?
Although using rust-lang/rfcs#1399 would be better, I would like this merged in the meantime to fix a problem I am having.

retep998:
I possibly have a better fix in https://github.com/retep998/winapi-rs/commit/a06b9bc65019dcfb257d65595c1b43eded972a2c

mvdnes:
That works, and is a better solution :+1:

PR #279, opened 2016-05-07
Add basic ROAPI and HSTRING functions required to use WinRT APIs

contextfree:


lib/runtimeobject/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "runtimeobject-sys"
-version = "0.0.1"
+version = "0.1.0"
```
retep998: Don't change the version number. I do that when I publish a new version to crates.io

Boddlnagg:
This is a strict subset of #281

retep998:
#281 landed and since this is supposedly a strict subset, closing.

PR #281, opened 2016-05-15
Added many bindings for runtimeobject

Boddlnagg:
As the title says. This is required for working with the Windows Runtime.

Note that there's currently a [bug](https://sourceforge.net/p/mingw-w64/bugs/540/) in the bundled runtimeobject.a, so the linker can't link to `RoUninitialize` on the i686-pc-windows-gnu toolchain.

**Update:** This is now a strict superset of #279


src/roerrorapi.rs
```diff
@@ -0,0 +1,11 @@
+FLAGS!{enum RO_ERROR_REPORTING_FLAGS {
+    RO_ERROR_REPORTING_NONE               = 0x00000000,
```
retep998: No vertical alignment please.

contextfree:
This looks like nearly a superset of my #279  which adds just a handful of ROAPI and HSTRING functions, with the exception of exporting the RIDL and ENUM macros - I wanted those exported so I could define WinRT interfaces/types outside of the winapi crates (because bindings to those interfaces can/should ideally be generated from metadata, it doesn't really feel to me like they belong here). So if this is merged it should obsolete #279 except for the macro exports.

retep998:
I'm just gonna take a moment to point out that both of your automatically generated avatars are pink.

Boddlnagg:
Ohhh, I haven't seen your PR, or else I might have just used it. Looks like we have similar intentions, namely binding to WinRT, and autogenerating the interface definitions from metadata (winmd). Maybe we can somehow join forces to work on this?
Exporting the RIDL macro as you have done seems like a good idea, I can add this to my PR.

contextfree:
thanks. yeah I'd definitely be interested in collaborating on this, though I have to warn you I'm new to collaborating on distributed open-source projects ( #279 was my first ever GitHub PR) so I might still suffer from some "social awkwardness" here. :)  

Do you have any working example code that uses this stuff yet? I've gotten as far as getting one piece of awful test code that exercises an async WinRT API (including the necessary COM interface implementation for a completion handler) to "works on my machine" status. I set up a repository -> https://github.com/contextfree/winrt-rust but I haven't uploaded that code yet because it still contains local paths. My plan was/is to first try to refactor that example to be as less-awful as I can, so I have a better target for what to generate, then start work on the autogenerator (which I'm thinking of as a "low-level projection" analogous to WRL, with the idea that a "high-level projection" analogous to /CX or the C# or JS bindings could be added later. I'm not entirely sure if that's the right mental framework though.) The areas of "awfulness reduction" I'm targeting are HSTRING management, statics, and async invocation, so the result of this phase would be a helper library for those + my example refactored to use them. But if you've been working on similar stuff maybe you already have some of that and/or different ideas.

Boddlnagg:
So this really sounds a lot like I was planning, with the one exception that I named my "playground" project winrt-rs instead of winrt-rust and haven't created a GitHub project yet :wink: By the way, don't worry about social awkwardness!

I have created a very simple HSTRING abstraction, which works, but can be improved for sure, see this [gist](https://gist.github.com/Boddlnagg/ada2a3c3227e3c922f326b2c6580d58d) and feel free to use it. I also reused the `ComPtr` defined in https://github.com/retep998/wio-rs/blob/master/src/com.rs

I managed to get some basic instance creation and function calling to work, but am currently struggling with the problem that I want to call Windows.Devices.Enumeration.DeviceInformation.FindAllAsync(...).GetResult(), which results in an E_NOTIMPL. However, the same call from a C# Win32 project that imports Windows.Runtime also results in a InvalidOperationException, so there's something else going on here that has nothing to do with the Rust stuff. (Maybe I'm starting the runtime in a wrong way?)

I had looked into defining a completion handler, but haven't got anything working yet, so you seem to be a step ahead here, which is great :+1:  I haven't put any work into the automatic generation from metadata, besides having looked into Mono.Cecil (I wanted to do it in C# and maybe later port it to Rust).

Now we should probably take this discussion out of this PR thread, maybe into a meta issue in your repository?

Boddlnagg:
I added the macro exports, so this should now be a strict superset of #279

retep998:
Could you rebase and squash this PR?

Boddlnagg:
When I find out how to do that ... I've never done it before. Let me try.

Boddlnagg:
Squashed and rebased.

Boddlnagg:
What is this mysterious **continuous-integration/appveyor/branch** where you can't even see any progress/details?

retep998:
@Boddlnagg That is for myself. I always merge PRs into `dev` first, and then once appveyor passes on that I can fast forward it to `master`.

PR #283, opened 2016-05-16
Add 32-bit {Get,Set}{Class,Window}LongPtr{A,W} functions as re-exports.

Daggerbot:
The C headers define these functions as macros instead of functions when compiling for 32-bit. This adds these as re-exports on i686/x86.


pravic:
This resolves #285 also.

retep998:
In winapi 0.3 these functions now have re-exports.

PR #291, opened 2016-06-08
Add function for read version information from executable files

bozaro:
* GetFileVersionInfoSize
 * GetFileVersionInfo
 * VerQueryValue


lib/version/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "version-sys"
-version = "0.0.1"
+version = "0.0.2"
```
retep998: Don't change the version number. I do that myself when I publish a new version to crates.io

bozaro:
I reverted my changes from Cargo.toml

PR #292, opened 2016-06-11
Add .idea/ and *.iml to .gitignore

mmitteregger:
Add IntelliJ project files to .gitignore.
Useful when developing with the [intellij-rust plugin](https://github.com/intellij-rust/intellij-rust).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target/
 Cargo.lock
+.idea/
+*.iml


retep998:
Any files like these which are specific to a platform or your IDE of choice really should belong in your own personal gitignore. A project's gitignore should only cover temporary files created by its own build systems.

PR #294, opened 2016-06-14
Update README.md example to be target specific

test9753:
Without target specific dependencies, Cargo will try to download and compile windows specific crates on other targets (for e.g.: unix).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,10 @@ winmm-sys = "0.1"
 ```
 example.rs:
 ```Rust
-
 #[cfg(windows)] extern crate winapi;
 #[cfg(windows)] extern crate winmm;
+#[cfg(windows)]
 fn func() {
-    #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


retep998:
Winapi compiles fine on linux, so making the dependencies target specific is optional. Also, even though you made the `extern crate` things target specific, you didn't make the function itself target specific. It would probably be better to replace this example with something that calls `MessageBoxW` as it would give a demonstration of how to work with wide strings.

retep998:
I updated the example in the readme to use `MessageBoxW`. If you still want to go through with this PR feel free to rebase it on top of `dev`.

https://github.com/retep998/winapi-rs/commit/2fe8dc20134bb8874d3a058a3f5f9206c3692e71

retep998:
Closing PR due to inactivity but feel free to open a new PR if you want to pursue this still.

PR #304, opened 2016-07-15
Add partial wininet bindings.

DanielKeep:
Includes what _should_ be all the types and constants from wininet.h,
and some of the functions.

Note that, unlike other header bindings, the `wininet` module is _not_
re-exported in the root `winapi` module.  This is because `wininet.h`
and `winhttp.h` define the same constants/enums with _different values._


DanielKeep:
**Do not merge yet**: I haven't checked that structs respect packing.

retep998:
This PR will need to be updated to the winapi 0.3 new world order. https://github.com/retep998/winapi-rs/issues/316

DanielKeep:
PR has been updated.

The one thing I'm not sure about is that I've had to add an `Align` field to `GOPHER_ATTRIBUTE_TYPE` to get the appropriate alignment on 64-bit.  Specifically, it's a field that users won't expect to exist, and if better padding control comes along later, can't be removed.  Should I rename it, add some docs for it, or just comment it out along with everything that depends on it?

I mean, it's for Gopher, so I'm not sure anyone'll miss it.  Well, right away; you know how people are.

retep998:
I'd prefer leaving the struct in with insufficient alignment for now, and commenting out the test, and having a fixme to fix it later.

src/um/wininet.rs
```diff
@@ -0,0 +1,2323 @@
...
+pub const SECURITY_FLAG_IGNORE_CERT_DATE_INVALID: DWORD = INTERNET_FLAG_IGNORE_CERT_DATE_INVALID;
+pub const SECURITY_FLAG_IGNORE_REDIRECT_TO_HTTPS: DWORD = INTERNET_FLAG_IGNORE_REDIRECT_TO_HTTPS;
+pub const SECURITY_FLAG_IGNORE_REDIRECT_TO_HTTP: DWORD = INTERNET_FLAG_IGNORE_REDIRECT_TO_HTTP;
+pub const SECURITY_SET_MASK: DWORD = SECURITY_FLAG_IGNORE_REVOCATION
+    | SECURITY_FLAG_IGNORE_UNKNOWN_CA | SECURITY_FLAG_IGNORE_CERT_CN_INVALID
+    | SECURITY_FLAG_IGNORE_CERT_DATE_INVALID | SECURITY_FLAG_IGNORE_WRONG_USAGE;
+pub const AUTODIAL_MODE_NEVER: DWORD = 1;
+pub const AUTODIAL_MODE_ALWAYS: DWORD = 2;
+pub const AUTODIAL_MODE_NO_NETWORK_PRESENT: DWORD = 4;
+pub type INTERNET_STATUS_CALLBACK = Option<unsafe extern "system" fn(
```
retep998: We now have `FN!` for this.

DanielKeep:
Updated fixes.  From commit:

- Changed `!0` to `-1i32 as u32`.
- Removed redundant `#[allow(dead_code)]`s.
- Changed `GOPHER_ATTRIBUTE_TYPE` to be naturally aligned correctly.
- Changed pointer type aliases to use `FN!`.

Note that the vtable struct still uses literal types, since there's no
name defined for them.

retep998:
Merge conflicts and CI is failing.

DanielKeep:
My bad; I think I forgot to enable features when testing the fixes I made.  And *then* I forgot that `dev` was the upstream branch, rebased on to master, amended the commit, and *then* realised what I'd just done and had to fix it.

Looks good now.  I hope.

Cargo.toml
```diff
@@ -74,8 +74,11 @@ headers = ["headers-shared", "headers-um", "headers-vc", "headers-winrt"]
     "um-wincon",
     "um-winevt",
     "um-wingdi",
+    "um-wininet",
+    "um-winineti",
     "um-winioctl",
     "um-winnt",
+    "um-winnt",
```
DanielKeep: I blame git gui for this.

PR #313, opened 2016-08-08
Adding LookupAccountNameA and LookupAccountNameW

norbert78:
New pull request based on the (hopefully) correct master branch

### DIFF
--- a/lib/advapi32/src/lib.rs
+++ b/lib/advapi32/src/lib.rs
@@ -498,8 +498,12 @@ extern "system" {
     // pub fn LogonUserExExW();
     // pub fn LogonUserExW();
     // pub fn LogonUserW();
-    // pub fn LookupAccountNameA();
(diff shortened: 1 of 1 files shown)


retep998:
CI is failing

norbert78:
OK, just had a look and the issue seems to be about the use of LPCTSTR. 
I would guess that I would have to use LPCSTR or LPCWSTR instead. Is this correct? If this is the case, I would try to re-submit....

retep998:
Yes, you probably want to use `LPCSTR` instead of `LPCTSTR`. Also you need to indent the arguments inside the functions, and you're going over the line length limit.

You don't have to submit a new PR. Instead you can just amend your current commit and then force push it.

norbert78:
Thanks for the advice! I'm still learning :-)
I just issued another commit. I hope that CI passes now.

Issue #316, opened 2016-08-13
winapi 0.3 overhaul


retep998:
All progress is in the [dev](https://github.com/retep998/winapi-rs/tree/dev) branch.
- [x] Remove all trait impls except `Copy` and `Clone`.
  - [x] Update all macros to do so.
  - [x] Get rid of any derives that are floating around, not behind a macro.
- [x] Make enums nothing more than integer constants with the enum type being an alias for `u32`.
- [x] Update all copyrights and license stuff. https://github.com/retep998/winapi-rs/issues/308
- [x] For each header do the following:
  - Move it to the appropriate directory based on where it is located in the Windows SDK such as `um` or `shared`.
  - Create a feature for that header such as `foo`.
  - Add the appropriate `pub mod foo;` to that folder's module.
  - Add a cfg to that mod such as `#[cfg(feature = "foo")]`.
  - Get rid of all `::` and add imports to pull in any definitions it needs from other headers.
  - Add the feature to `build.rs` and specify its dependencies on other headers and also libraries.
- [x] For each -sys crate do the following:
  - Move the function definitions to the header module where it is actually defined.
  - Ensure that any header that has functions from that library specifies that dependency in `build.rs`.
- [x] Update all usage of `UNION!` to `UNION2!`. Once that is done, rename `UNION2!` to `UNION!`.
- [x] Inform everyone with an open PR that they'll need to update their PR for the new world order.

alexcrichton:
One change I might also recommend is `#![no_std]` by default if you can get away with it, libc right now only does so because it was stable before `#![no_std]` was stable. Other than that though this sounds great to me!

Once the refactoring is done, perhaps some measurements could be done to ensure that issues like https://github.com/retep998/winapi-rs/issues/218 are addressed as well? Seems like it'll definitely get smoothed over regardless.

retep998:
@alexcrichton Unless drastic improvements are made in rustc to improve compile times, I don't think winapi will ever be able to compile quickly. Just `winnt` alone (needed for basic things like `HANDLE`) takes a couple seconds. This is already with all trait impls gone except `Copy` and `Clone` and enums replaced with mere integer constants, so there's no way to make it any faster aside from simply not including a lot of code.
```
$cargo build --features shared.winnt
   Compiling winapi v0.3.0-dev (file:///C:/msys64/home/Peter/winapi-rs)
    Finished debug [unoptimized + debuginfo] target(s) in 2.36 secs
```
[...]

Issue #302, opened 2016-07-14
Create .def files so I can have my own import libraries


retep998:
Since it is actually really easy to create import libraries from `.def` files which are plain text files that can be easily edited in a text editor, I should get my own set of `.def` files for all system libraries that are up to date with the latest Windows 10 version. This would allow for the creation of high quality up to date import libraries for MinGW, so I don't rely on the outdated and sometimes buggy import libraries from MinGW.

As a future extension to this, I can also create MSVC import libraries from these `.def` files, which can be used on non-Windows platforms that would like to target Windows using the `-pc-windows-msvc` targets using `lld` with the link flavor. This would allow the highest quality cross compiling to Windows with no MinGW cruft and no proprietary blobs.

The repository would contain only the `.def` files with a script to create import libraries from them. The import libraries would in turn be distributed through crates on crates.io which `winapi` would depend on conditionally using target specific dependencies. There would be a separate crate for each target, so that the user only downloads the version that they need.

retep998:
Creating a `.def` file for stdcall symbols on Windows 32bit is actually incredibly difficult. I'd probably have to write my own `.dll` with stubs for all the functions along with the `.def` and then use the resulting `.lib`, unless someone has a better idea.

retep998:
With MinGW it is doable apparently with `dlltool`. stdcall symbols need to be defined as `foo@N` in the def files and then the import libraries created by doing `dlltool -d foo.def -l libfoo.a -k`. It shouldn't be too hard to whip up a program to dump the symbols from Microsoft's import libraries and create def files from them.

retep998:
This has now been solved! I wrote a [fancy import library generator](https://github.com/retep998/windump-rs/blob/master/src/bin/exports.rs).
I then generated a bunch of import libraries and checked them into the repo, along with their .def files! https://github.com/retep998/winapi-rs/commit/1f932ceeb5968d7b9f75311a5b019128c3c2f261 and https://github.com/retep998/winapi-rs/commit/7eb10e7c9c4392bdd5c306dccb8ad78b5cf4f8c7

PR #324, opened 2016-08-23
implements avrt

k-brac:
Implements avrt
Adds a macro for enum having negative values


retep998:
With the `-gnu` targets Rust is able to work standalone without a separate C/C++ toolchain. This means that the only import libraries which are available are either those that Rust bundles, or winapi bundles. In the case of avrt, Rust does not bundle it, therefore winapi _must_ bundle `libavrt.a` itself. The current way of resolving this is getting `libavrt.a` from a mingw-w64 installation and putting it in the correct location. See https://github.com/retep998/winapi-rs/tree/master/lib/secur32 for an example of this. However in the near future I will likely be solving this with https://github.com/retep998/winapi-rs/issues/302 instead.

src/avrt.rs
```diff
@@ -0,0 +1,17 @@
...
+
+ENUM_NEG!{enum AVRT_PRIORITY {
+    AVRT_PRIORITY_VERYLOW = -2,
+    AVRT_PRIORITY_LOW,
+    AVRT_PRIORITY_NORMAL,
+    AVRT_PRIORITY_HIGH,
+    AVRT_PRIORITY_CRITICAL,
+}}
+
+ENUM!{enum AVRT_NOTIFICATION {
```
retep998: Strangely enough, `AVRT_NOTIFICATION` was actually removed in the Windows 10 SDK. Since I try to track the latest Windows 10 SDK with winapi, I'd rather this not be here, unless you can provide a compelling reason for why it should be kept.

k-brac: My bad. My visual studio was not properly setup to use the latest sdk... Hence I removed AVRT_NOTIFICATION
The rest doesn't seem to have been modified

retep998:
This PR will need to be updated for the winapi 0.3 new world order. https://github.com/retep998/winapi-rs/issues/316

k-brac:
merged in #348

PR #330, opened 2016-09-09
Define FOREGROUND_BLUE, etc. as WORD

oconnor0:
Define FOREGROUND_BLUE, etc. as WORD since they are intended to be used to create Attributes which are also defined as WORD - as in https://msdn.microsoft.com/en-us/library/windows/desktop/ms682013(v=vs.85).aspx


retep998:
This was merged without your merge commit, so github didn't recognize it as having merged the PR.

PR #334, opened 2016-10-02
Remove derives outside of macros

angelsl:


retep998:
All of these `IFoo` things are actually COM interfaces that need to be fleshed out. They should probably just be deleted until someone fills them in properly.

angelsl:
> should probably just be deleted

So, should I do it?

PR #338, opened 2016-11-05
Updated various modules to 0.3

brashaw:
Refer to the commits.

PR #340, opened 2016-11-14
Add handleapi header to 0.3

trlim:
This adds a definition of `INVALID_HANDLE_VALUE` from `um/handleapi.h`
which is the only definition in the header.
It also removes the same old definition from `shlobj.rs`.


retep998:
Actually `um\handleapi.h` does contain a bunch of function definitions.

trlim:
You're right. I had to say "the only **non-function** definition".

I didn't give much thought on those function definitions because they(except `CompareObjectHandles`) are already defined in `lib/kernel32` but I should have read your comment in #338 more carefully.

If my understanding is correct, at least the following changes required.
- [ ] Move those function defs from `kernel32` to `um/handleapi` header
- [ ] Make `kernel32` library to be a dependency of `um/handleapi`

retep998:
Yep, seems correct.

PR #341, opened 2016-11-16
Add bindings for `setupapi`

trlim:
`setupapi` and its dependent header `spapidef` are updated for 0.3.
This also includes `devguid` which is required to use some `setupapi` functions.


trlim:
Don't know why travis fails to build for nightly.

retep998:
Travis failure is due to https://github.com/rust-lang/rust/issues/37786

retep998:
Appveyor failure is due to https://github.com/rust-lang/cargo/issues/3384

trlim:
Found some function signatures are wrong and realized some definitions are not fully following guidelines.
I added two commits to fix that.

PR #347, opened 2016-11-29
Defined IShellItemArray in shobjidl.rs

Raymai97:
On top of that, I also added the definition of types required by this interface.
Test OK when using code like `IFileOpenDialog.GetResults(...)`.


retep998:
Merged into `dev` branch.

PR #351, opened 2016-12-09
Update `objbase`,  `combaseapi` and `objidl` for 0.3

trlim:
This PR moves 0.2 bindings of `objbase`, `objidl` and `combaseapi` to 0.3 compliant modules.
It doesn't try to add new bindings except for some enums in `combaseapi`.


trlim:
Every IDL generated interface definition also has IID definition like `EXTERN_C const IID IID_IBindCtx;` that its binding is defined in `uuid` crate.
Should I also move this to its header module(`objidl` in this case)?
I think I also have to add `uuid` library feature and make `um/objidl` to be dependent on it in that case.

retep998:
What I'm actually going to be doing is adding a way to specify the IID of each interace in the `RIDL!` invocation itself. I just haven't gotten around to doing so yet.

trlim:
That's a good idea!
I think we will also need a macro to specify CLSID.
And eventually, I hope something similar to `__uuidof` in MSVC to be possible in the future.

retep998:
PR has conflicts.

retep998:
PR has conflicts. Also `#[uuid]` was added so you'll need that too.

trlim:
Rebase merged and fixed conflicts.

retep998:
PR has conflicts again.

PR #365, opened 2017-01-20
Add categories to Cargo.toml

shepmaster:
Hi! [crates.io now supports categories][categories], which are a curated list
of topics aimed at helping an end-user coming to crates.io looking for
"a crate to do ______".

We're sending pull requests to selected crates to add categories in order to help 
populate the categories and seed their usefulness. We've made a guess at the best
category/categories for this crate; if it doesn't fit, please feel free to take
a look at [all the available categories and their descriptions][categories] and
[the slug values that should be specified in your Cargo.toml][slugs] and pick
different ones. If you have a category in mind that isn't available, you can
[send a PR to this file on crates.io][categories.toml] to propose additional
categories.

Crates can have up to 5 categories, and uploading categories to crates.io
currently requires publishing a new version with a cargo nightly from 2017-01-18
or later (it needs to contain [this PR][cargo-pr]).

We've [published a blog post][blog-post] with further details about categories.
The blog post also talks about the new crates.io support for CI badges, which
you may be interested in adding as well.

Please let me know if you have any questions!

[categories]: https://crates.io/categories
[slugs]: https://crates.io/category_slugs
[categories.toml]: https://github.com/rust-lang/crates.io/blob/master/src/categories.toml
[cargo-pr]: https://github.com/rust-lang/cargo/pull/3301
[blog-post]: http://integer32.com/2017/01/20/categories-and-ci-badges.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ documentation = "https://retep998.github.io/doc/winapi/"
 repository = "https://github.com/retep998/winapi-rs"
 readme = "README.md"
 keywords = ["windows", "ffi", "win32", "com", "directx"]
+categories = ["os::windows-apis"]
 license = "MIT/Apache-2.0"
 include = ["src/**/*", "Cargo.toml", "LICENSE-MIT", "LICENSE-APACHE", "build.rs", "README.md"]
 build = "build.rs"


retep998:
Would `external-ffi-bindings` also be applicable?

shepmaster:
> Would `external-ffi-bindings` also be applicable?

That's a great question, but I don't know enough about the winapi crate to say for sure. 

The general idea is that `*-sys` crates should definitely be `external-ffi-bindings`. I'm hoping you have a magic way to apply changes to the great number of the `*-sys` packages in the Windows area in an efficient manner as I don't know if my fingers could handle it 😇 .

The other possibility is `api-bindings`, which would be better suited for a layer on top of the raw bindings.

One example is `openssl` (`api-bindings`) and `openssl-sys` (`external-ffi-bindings`).

If winapi provides a more Rust-like interface, then I'd say `api-bindings`, or if it just offers up a more direct binding, I'd go with `external-ffi-bindings`.

retep998:
winapi does not offer any sort of Rust-like interface. It is purely raw bindings, so `external-ffi-bindings` is probably right.

As for the hundreds of `-sys` crates, I abandoned that approach in winapi 0.3 due to various reasons.

retep998:
Categories added in https://github.com/retep998/winapi-rs/commit/81de8e3ce196af4ec54c6b8518405d8b7a509cb3

PR #368, opened 2017-02-04
Add several control constants to winuser.rs

Osspial:


retep998:
CI is failing

Osspial:
Hmm, looks like I didn't have the proper feature flag enabled and my additional code wasn't actually compiling on my machine! Anyway, there were two notification codes that I tried to add and wasn't entirely what to do about: [`LBN_ERRSPACE`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb775155(v=vs.85).aspx) and [`CBN_ERRSPACE`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb775816(v=vs.85).aspx). Both of those codes are represented by negative numbers (`-2` and `-1`, respectively), but the MSDN indicates that they're supposed to be represented as unsigned `WORD`s. I've commented them out for now, but some sort of solution about what they're actually represented as should probably be found.

retep998:
`-1i16 as u16`

PR #374, opened 2017-02-19
Update libraries for 0.3

schulzch:
This PR does:
- Parse list of libraries from `Cargo.toml` (replaces `LIBS` in `build.rs`).
- Adds more libraries, based on directory names in `lib`

Should be in line with #316.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ license = "MIT/Apache-2.0"
 include = ["src/**/*", "Cargo.toml", "LICENSE-MIT", "LICENSE-APACHE", "build.rs", "README.md"]
 build = "build.rs"
 
+[build-dependencies]
+toml = "0.3"
+
 [target.i686-pc-windows-gnu.dependencies]
 winapi-i686-pc-windows-gnu = { version = "0.3", path = "i686" }
 [target.x86_64-pc-windows-gnu.dependencies]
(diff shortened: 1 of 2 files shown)


schulzch:
CI errors are weird (missing libraries on gcc and unknown error for rust 1.0)

retep998:
`toml` depends on `serde` which I believe at some point released a version which could not be parsed by older Cargo. Therefore `toml` would significantly raise the minimum Rust requirement for winapi which I do not want to do at this time. Also I don't like depending on external crates anyway (they'll steal all my downloads!).

schulzch:
Ok, I'll remove that part.

I've been playing around with using `bindgen` on `windows.h` and it works okay-ish, except for cast-style constants like `#define SOME_CONST ((int) 3)`. I'll push it later on. What are your thoughts on automation in general?

retep998:
I don't mind people using bindgen to help them write winapi bindings as long as the end result is accurate and in the correct style.

schulzch:
Done. CI issue seems to be related to AppVeyor (missing libraries).

In case you are interested in experiments with ´rustgen` see [winapi-gen-rs](https://github.com/schulzch/winapi-gen-rs).

retep998:
PR has rather major conflicts.

retep998:
This PR basically has no point anymore given the way that library dependencies are now handled.

PR #376, opened 2017-02-19
add LPSTR variants

BlackYoup:


retep998:
Winapi does not add definitions which depend on `UNICODE`. You must explicitly use either the W or A versions of functions as necessary (hopefully always W and never A).

PR #406, opened 2017-03-20
Add oaidl::ITypeLib, oleauto::LoadTypeLibEx and related definitions.

Arnavion:
There's more in oleauto.h but I've only added what I need for winapi-tlb-bindgen.

There's some amount of duplication between oaidl.h and oleauto.h (both define `DISPID` and `MEMBERID`, etc) but I figured it would be better to reuse the definitions from the `oaidl` module.

`MEMBERID_NIL`'s type comes from its use as the first parameter of [`ITypeInfo::GetDocumentation()`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms221227)


src/um/oleauto.rs
```diff
@@ -0,0 +1,34 @@
...
+use um::oaidl::{DISPID_UNKNOWN, ITypeLib, MEMBERID};
+use um::winnt::{HRESULT, INT};
+
+EXTERN!{stdcall fn SysAllocString(
+    psz: *const OLECHAR
+) -> BSTR}
+
+EXTERN!{stdcall fn SysFreeString(
+    bstrString: BSTR
+) -> ()}
```
Arnavion: Is this correct or should it be c_void or something?

retep998: `void` return types correspond to `()` in Rust.

src/um/oaidl.rs
```diff
@@ -526,7 +526,53 @@ interface IDispatch(IDispatchVtbl): IUnknown(IUnknownVtbl) {
 );
 pub enum IRecordInfo {} // FIXME
 pub enum ITypeComp {} // FIXME
-pub enum ITypeLib {} // FIXME
+
+ENUM! { enum SYSKIND {
```
retep998: Incorrect spacing.

src/um/oleauto.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright © 2016-2017 winapi-rs developers
```
retep998: Shouldn't this be just 2017?

retep998:
If there is duplication the definitions should be duplicated except for structs and interfaces, where one location should be canonical and then the other headers publicly re-export it.

Arnavion:
>If there is duplication the definitions should be duplicated

Okay, changed it to redeclare `DISPID` and `MEMBERID` instead of importing the latter from `oaidl`

src/um/oleauto.rs
Arnavion: Fixed.

src/um/oaidl.rs
Arnavion: Fixed.

Issue #399, opened 2017-03-16
Add support for isolation aware mode


Osspial:
The Windows API has an "isolation-aware" mode, where several function calls are re-defined to create and reference data specific to a submodule (i.e. dlls or dependent crates), instead of sharing data with the host executable. This is done by adding `IsolationAware` to the start of the function symbol (e.g. `DefSubclassProc` being replaced with `IsolationAwareDefSubclassProc`). Doing this would allow dependent crates to define their own application manifests, without them relying on the root executable to define the manifest. These re-definitions seem to be found in the `{headername}.inl` files inside the Windows SDK `um` folder, with most of them being in `commctrl.inl` (although there are a few others).

I think that `winapi` should provide support for these functions, although it looks like it would be a pretty huge undertaking. Relevant MSDN documentation for these can be found [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa375197(v=vs.85).aspx) and [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms649781(v=vs.85).aspx#extensions).

retep998:
Isolation awareness would be a huge undertaking and I have no plans to work on it for a long time. I'm not opposed to accepting a PR for this however.

PR #422, opened 2017-03-28
Add isolation aware support to 0.2 branch

Osspial:
Adds macros to the `kernel32`, `user32`, `comctl32`, and `comdlg32` libraries on the `0.2` branch to create isolation-aware functions. These are exposed as macros rather than as modules in the respective crates because isolation aware mode requires library-local data to be created in order to hold the activation context, and macros let the data be inlined into the using library. Right now, this doesn't handle cleaning up the created activation contexts, though I'm not entirely sure how to expose that functionality.

This was developed against the `0.2` branch because I needed to use it for my own crate, but porting it to `0.3` should just be a matter of moving the `isolation_aware.rs` files and changing a few imports.


Osspial:
Would close issue #399

Susurrus:
I believe that @retep998 is only accepting PRs to the active development branch `dev`, so using the 0.3 API. Could you rebase this on that branch instead? You can set the merge branch to `dev` by hitting the `Edit` button next to the PR title.

retep998:
Reviewing this PR is being postponed until winapi 0.3 is finally published, at which point I will investigate things like these.

nico-abram:
0.3 is released right? Surprisingly, I don't see conflicts according to github

So this is review-able? AIUI, there's a lack of reviewers for winapi atm, so maybe it'd be easier to get this merged in chunks as separate stand alone PRs?

PR #431, opened 2017-04-08
Move all kernel32 fileapi functions to fileapi

Susurrus:
Not positive about the changes to build.rs, but I think they're right.


retep998:
Could you format the functions to be block indented with one parameter per line?

Susurrus:
Okay, I fixed those things and rebased. Hopefully it looks good now. I'll do the same to the others as well.

retep998:
Could you linebreak even the functions that can fit on a single line? The only ones that should be a single line are the ones with no parameters at all.

Susurrus:
All functions now have single-line parameters. Could this requirement be specified in  CONTRIBUTING.md? Structs are specified there to always be 1 field per line, but the 1-parameter-per-line requirement isn't specified.

PR #433, opened 2017-04-08
Migrate commapi stuff out of kernel32

Susurrus:
These functions are also declared in winbase, but I didn't export them there as well until I figured out if I was on the right track with this.


Susurrus:
Rebased this, though there aren't any style issues that I saw here, so this should be good to merge.

Susurrus:
This has also been corrected to use 1-parameter-per-line function declarations.

PR #437, opened 2017-05-03
Move all winbase declarations out of kernel32 & advapi32

Susurrus:
Couple of notes on this:

* InterlockedCompareExchange64 doesn't exist in winbase.h even though it says that online, so I removed it.
* Interlocked* are all declared as C++, but they were in the original file. Should I have included them or should I drop them as well? They have multiple declarations in the original file, so I wasn't sure which to use.
* Don't know what to do with PIPE_ATTRIBUTE_TYPE since it nor any functions that depend on it are declared in any of the Windows SDK headers, so I removed it.
* Two enums were moved to `fileapi` and `processthreadsapi` and now winbase depends on them and the kernel32 lib

I compiled this locally on Linux, but it doesn't compile cleanly just specifying a target, so I'm relying on CI testing for testing the feature & library dependencies. I haven't resolved the import situation yet since I don't have a Windows machine to test on, so we'll see how bad it is in CI.


retep998:
`InterlockedCompareExchange64` is defined in `winnt.h` though.

Susurrus:
Re `InterlockedCompareExchange64` a comment in `winnt.h` says:
  > Skip interlocked and bit manipulation stuff because it is all intrinsics, use the native Rust equivalents instead

So it's fine to remove this function?

Susurrus:
This doesn't use the `UNION2!` macro, but otherwise this passes all tests and I think it should be sound.

retep998:
If something is an intrinsic or a wrapper around an intrinsic, then you can just leave a comment indicating what would be there such as `// InterlockedCompareExchange64`.

Susurrus:
Then should I remove all the `Interlocked*()` functions from WinBase and instead make them those comments. There were some already implemented and some weren't.

retep998:
If it is actually defined in the headers as an inline function that wraps an intrinsic, then that can't be done in winapi so it would be left out. If it is defined in the headers as an external function that you can link against, then it should be included.

Susurrus:
I don't know how to check if they're intrinsics, by just looking at the definitions, but they're all inline. For example:
```
FORCEINLINE
unsigned long
InterlockedIncrement(
    _Inout_ _Interlocked_operand_ unsigned long volatile *Addend
    )
{
    return (unsigned long) _InterlockedIncrement((volatile long*) Addend);
}
```

This looks similar for all of them except the `*64()` ones, where the inner function call name doesn't have a leading underscore. So I'm not certain which of these are intrinsics.

retep998:
`_InterlockedIncrement` is definitely an intrinsic because `#pragma intrinsic(_InterlockedIncrement)` and `InterlockedIncrement` is just a wrapper around it. In fact, of all the functions that start with `Interlocked`, the only functions which are even in the import libraries are...
```
InterlockedFlushSList
InterlockedPopEntrySList
InterlockedPushEntrySList
InterlockedPushListSList
InterlockedPushListSListEx
```
All the other interlocked stuff is simply intrinsics.

Susurrus:
So I think that means I can just ignore all of them (make them into commented-out function names), yes?

Susurrus:
I've commented out the intrinsics. Should be good to go at this point.

src/um/winbase.rs
```diff
@@ -326,7 +1056,615 @@ pub const STARTF_TITLEISLINKNAME: DWORD = 0x00000800;
...
+        lpTemplateDirectory: LPCWSTR, lpNewDirectory: LPCWSTR,
+        lpSecurityAttributes: LPSECURITY_ATTRIBUTES,
+    ) -> BOOL;
+    pub fn CreateDirectoryTransactedA(
+        lpTemplateDirectory: LPCSTR, lpNewDirectory: LPCSTR,
+        lpSecurityAttributes: LPSECURITY_ATTRIBUTES, hTransaction: HANDLE,
+    ) -> BOOL;
+    pub fn CreateDirectoryTransactedW(
+        lpTemplateDirectory: LPCWSTR, lpNewDirectory: LPCWSTR,
+        lpSecurityAttributes: LPSECURITY_ATTRIBUTES, hTransaction: HANDLE,
```
retep998: You forgot to format these functions.

Susurrus: Fixed.

PR #439, opened 2017-05-05
Add rustfmt config file

Susurrus:
This follows the winapi-rs-specific style guidelines, but it doesn't work properly for functions with 1 argument, so not currently stable for running on the whole repo, but this would assist with porting. I've already filed a bug with upstream (rust-lang-nursery/rustfmt#1508).

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,4 @@
+fn_args_density = "Vertical"
+fn_args_layout = "Block"
+fn_call_width = 0
+trailing_comma = "Always"


Susurrus:
Note that rustfmt doesn't allow keeping binary operators at the beginning of a new line when they've been broken. I'd suggest using the default rustfmt here to be compatible with the rest of the rust ecosystem or we should file a bug to add this to rustfmt, though I don't know if they'd implement it.

Susurrus:
Closing this in favor of a tracking issue where this can all be discussed.

PR #440, opened 2017-05-07
Import tests

GuillaumeGomez:
Include another PR of @Susurrus.


Susurrus:
Sorry, I meant you should rebase your fixes off the `dev` branch and not include changes pertaining to my PR. I'll fix my PR with the appropriate changes. That way this is small and can be merged before my changes to make sure everything's in order.

GuillaumeGomez:
Hum... You don't make my life easy! :p

GuillaumeGomez:
Ok, so it doesn't include any other PR but fix a test and improve another one.

Susurrus:
Thanksm, @GuillaumeGomez! Sorry about the additional work, but hopefully this will be easier for @retep998 to merge.

It looks like there's a blacklist for headers to ignore. Can you add some comments in the code explaining why that exists and how to use it? We might want to have this documented in `CONTRIBUTING.md` as well so that if people try to add a library they will understand that adding it to this list might be a necessary step.

GuillaumeGomez:
In your case, it's because `sspi.rs` in "available" in two places. The one in the `um` folder is actually just a redirection to the one in `shared`, which is why it should be ignored.

And we created those tests in order to help people. When they add something, the tests will fail and will say exactly why. That's the whole point. :)

PR #445, opened 2017-05-12
Add d2d1effects_1.rs and d2d1_2.rs

Connicpu:


Susurrus:
LGTM. Correct declaration order and passes tests. As far as I can tell this is good to merge.

PR #438, opened 2017-05-04
Porting more APIs to 0.3

Susurrus:
This is blocking on #437, which is why that commit shows up here as well. I've realized merging can be problematic when pulling stuff out of these large older source files, so I've been stacking them up. Maybe I'll submit a bunch as a single PR in the future so the inter-patch dependencies are less of an issue.

I'm filing this now to get a head-start on any compilation builds on CI.

PR #446, opened 2017-05-20
Fix headers test

GuillaumeGomez:
Fixes #438.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,9 +5,8 @@
 // All files in the project carrying such notice may not be copied, modified, or distributed
 // except according to those terms.
 #![cfg(windows)]
-#![allow(bad_style)]
 #![deny(overflowing_literals, unused_qualifications, unused)]
-#![allow(unused_macros)]
+#![allow(bad_style, unused_macros)]
 #![recursion_limit = "128"]
 #![no_std]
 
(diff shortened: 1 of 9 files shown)


Susurrus:
@GuillaumeGomez These failed the tests on nightly, though I'm not certain why they failed like they did.

GuillaumeGomez:
Because these macros are only used by some features. I'll need to find which and turn them off for others.

Susurrus:
Maybe just allow unused macros?

Susurrus:
`#![allow(unused_macros)]` at the beginning of `src/macros.rs`?

GuillaumeGomez:
I don't like when things are too simple.

Susurrus:
I think you have to put that after the `deny(unused)`.

GuillaumeGomez:
Working!

Susurrus:
LGTM 👍

src/lib.rs
```diff
@@ -7,6 +7,7 @@
 #![cfg(windows)]
 #![allow(bad_style)]
```
retep998: Why couldn't you have added the `allow(unused_macros)` to this existing line?

GuillaumeGomez: Seems like it has to be declared after the deny for some reasons...

retep998: So then why not move it?

PR #450, opened 2017-05-22
Move all bcrypt functions to shared

kichristensen:


retep998:
There is no `bcrypt.h` in `um` in the Windows SDK.

kichristensen:
Sorry, it is in shared, I'll fix it

Susurrus:
@kichristensen Looks like you have `winnt` as a required feature but it's not actually used. Probably just need to remove it.

Susurrus:
@kichristensen Also, like your other commit, the function declaration order isn't correct. You should follow the same order as they are listed in the Windows header file.

Susurrus:
This is unnecessary now that #459 was merged, yes?

PR #453, opened 2017-05-30
Add new SetConsoleMode flags

mcgoo:
Add flags to allow enabling VT100 escape sequence processing in Windows consoles.


Susurrus:
All PRs should be against the `dev` branch (which will be winapi 0.3). These constants also already exist in the `src/um/wincon.rs` file.

mcgoo:
Got it, thanks - withdrawing the PR. Perhaps you should update the contribution guidelines to indicate that you only want PRs against `dev`?

retep998:
Really I should change it to be a tag instead of a branch.

PR #448, opened 2017-05-22
Porting more APIs to 0.3

Susurrus:
Re-submission of #438 since it was accidentally closed.

PR #454, opened 2017-05-30
More migration of src/*.rs

Susurrus:
Commits up to `Remove finished libs` are already in #448, so those can be ignored when reviewing this PR.


Susurrus:
@retep998 This should be good to go.

Susurrus:
This has now been rebased and I cleaned up a few more style issues that I noticed. This is ready for review.

PR #455, opened 2017-05-31
Finishing off the rest of src/*.rs

Susurrus:
After this it's only the source files in `lib/` that need to be ported.

98eab59 needs careful review since the datatype should actually be a `UNION2!`, but I'm not certain how to do that.

8a1d180 I also need some help with since I didn't figure out how to use `UNION2!` to port `sockaddr_in6`.

Note that this builds off of #448 and #454, which is why it looks so huge. This is actually only the commits since 2233a23.


Susurrus:
There are some problems with porting over some of the interfaces where they don't have a UUID in the header describing them. Specifically: `vswriter::IVssWriterComponents`, `vswriter::IVssWMFiledesc`, `vswriter::IVssWMDependency`, `vswriter::IVssWMDependency`, and `vsbackup::IVssWMComponent`.

Is there something I'm missing or should these be declared differently?

Susurrus:
This should now build as I'm using all-zeroes for missing UUIDs and I cleaned up some other issues.

Also, I redid `in6_addr` as a `UNION2`. I think it's correct, but I'd need confirmation on that.

Susurrus:
@retep998 Thanks!

PR #474, opened 2017-07-10
Migrate more of lib/ to v0.3

Susurrus:
Gets rid of 6 more DLLs from `lib/`.


Susurrus:
Without any other PRs this brings the lines of code left in `/lib` to 6457 across 21 files.

Susurrus:
@retep998 It looks like MSVC builds are passing but GNU builds fail because of a missing `shcore` lib. I don't know where to start to debug that, would you be able to point me in a direction? All the docs seem to suggest it should work.

retep998:
`shcore` is not a simple import library for a single `.dll`. It actually contains symbols from multiple DLLs and I never got around to figuring out how to get MinGW to create that sort of library. If you look in `i686\lib` you'll see that `libshcore.a` is missing.

Susurrus:
So what should I do about this then? Drop the reference to `shcore` or just delete all the functions exposed by it for now?

retep998:
What I actually need is the correct `ar` incantation to merge multiple import libraries together.

Susurrus:
I meant what I should do for this PR. I don't know when you'll get around to looking at `shcore`, and it sounds like it might be a tricky thing to solve, so I'll just drop that commit from this PR so the rest of this is mergeable as-is at least.

Susurrus:
@retep998 Well this thing passes CI and looks good on my end. Would appreciate a review when you get a chance so I can move on to migrating more of these.

src/um/dsound.rs
```diff
@@ -185,3 +185,57 @@ pub const DSBPLAY_LOCSOFTWARE: DWORD = 0x00000004;
...
+    //     pcGuidRenderDevice: LPCGUID,
+    //     pcDSCBufferDesc: LPCDSCBUFFERDESC,
+    //     pcDSBufferDesc: LPCDSBUFFERDESC,
+    //     hWnd: HWND,
+    //     dwLevel: DWORD,
+    //     ppDSFD: *mut LPDIRECTSOUNDFULLDUPLEX,
+    //     ppDSCBuffer8: *mut LPDIRECTSOUNDCAPTUREBUFFER8,
+    //     ppDSBuffer8: *mut LPDIRECTSOUNDBUFFER8,
+    //     pUnkOuter: LPUNKNOWN,
+    // ) -> HRESULT;
```
retep998: Why are these commented out? Do they rely on a lot of types that haven't been defined yet?

Susurrus: Yes, all of these rely on structs that haven't been defined. Since they were like this in the original file, I decided to keep them in the new ported file.

Susurrus:
Wow, didn't realize you could push changes into my PRs, interesting. Looks like this still passes, are you currently reviewing this PR?

PR #476, opened 2017-07-11
Migrate shcore functions to 0.3

Susurrus:
This currently fails because of issues with `shcore` not being exposed as a library. @retep998 seems to understand the complete issue, so this PR is pending resolving that library.


retep998:
I've added the rest of the import libraries for MinGW targets: https://github.com/retep998/winapi-rs/commit/f628c80106e97516e8031880995727b7a14d6615

Susurrus:
Then let's run it again and see what we get!

Susurrus:
Hot-diggity-damn it worked!

Susurrus:
I ran it without rebasing and it worked, which is interesting. This one is properly rebased, but it should still work if the last one did.

retep998:
I believe on PRs appveyor actually tests the result of merging the PR onto the target branch.

Susurrus:
Interesting, didn't know that! I've used bors and homu to solve that problem, didn't realize it.

So does this look good to merge? This is a relatively small one here!

PR #466, opened 2017-07-03
Add winnt definitions

skdltmxn:

PR #497, opened 2017-09-11
Fill out memoryapi

docbrown:


docbrown:
Needs #466 first.

docbrown:
Test are failing with:

> Doc-tests winapi
error: An unknown error occurred

They pass on my machine, though. Hmmm...

retep998:
You have functions in `memoryapi` which come from a different library than `kernel32`. Specifically `MapViewOfFileNuma2` which comes from one of `onecore`/`onecoreuap`/`mincore`/`mmos`. In Rust 1.6 inline functions, such as `MapViewOfFile2`, are not lazily instantiated and `ld` fails to strip unreferenced functions so it fails to link. I'm not entirely sure why the doc test fails on current Rust but my guess is that it depends on a symbol which only exists on a newer version of Windows than Appveyor has.

retep998:
Doc tests failing because of https://github.com/rust-lang/rust/issues/44487

Susurrus:
@retep998 This good to merge now?

PR #505, opened 2017-09-30
Add d3d11sdklayers.h

siler:
Couple of oddities with this one:
  - Huge enumeration caused increase of macro recursion.
  - Some of these constant names are absurdly long and might break the 99 char line length limit.


retep998:
I agree, that is an absurdly huge enum with absurdly long names.

PR #507, opened 2017-10-03
Move WinNls functions from lib/kernel32

bbqsrc:
As per title.


build.rs
```diff
@@ -256,7 +256,7 @@ const DATA: &'static [(&'static str, &'static [&'static str], &'static [&'static
     ("winineti", &["minwindef"], &[]),
     ("winioctl", &["devpropdef", "minwindef", "winnt"], &[]),
     ("winnetwk", &["basetsd", "minwindef", "windef", "winerror", "winnt"], &["mpr"]),
-    ("winnls", &["basetsd", "guiddef", "minwindef", "winnt"], &[]),
+    ("winnls", &["basetsd", "guiddef", "minwinbase", "minwindef", "winnt"], &[]),
```
retep998: You need to add `"kernel32"` to the second array, because this module is linking in symbols from `kernel32.lib`.

PR #501, opened 2017-09-13
Add sysinfoapi definitions

skdltmxn:

PR #500, opened 2017-09-12
Finish threadpoolapiset

docbrown:

PR #447, opened 2017-05-21
Finish porting advapi32 to v0.3

austinwagner:
Used @Susurrus's fork as a base.

All functions exported by advapi32 that can be found in the SDK headers* have been added to their corresponding Rust files with the exception of SetServiceBits (um/LMServer.h). If a Rust version of a header already existed then only the required pieces were added to the file. If the Rust file didn't exist already the full header was ported to the extent I was able to.

*139 functions are exported from the DLL but have no declaration in any header in the latest Windows SDK.

PR #426, opened 2017-04-05
Update d3d12 SDK 10.0.14393.0 Remove d3d12-sys

red75prime:

PR #510, opened 2017-10-04
Finish migrating all 0.2 code in /lib to 0.3

Susurrus:
I'm submitting this as 1 big PR because it should a) be relatively easy to review since it's just migrating functions and there's no complicated data structures here and b) there are a ton of merge conflicts as soon as you start modifying `kernel32/mod.rs`, so this will be faster doing this in a big batch rather than having to ask people to update their PRs after every other one is merged.

After this there are only 348 lines of code left to migrate, most of which is covered by #447, #500, and #501.

The following should be noted:

- This PR contains only the migration pieces from #426 whose author said it might be a while before he can work on that PR again
- It also contains the same changes as done in #507, which was submitted after I made this change.


Susurrus:
@retep998 I was going to go ahead and basically finish kernel32 as part of this PR, or would you rather I not make this any bigger than it already is?

Susurrus:
Okay, the scope of this expanded a bit and this PR entails the porting of ALL functionality currently left in `/lib`. This means it supplants all the work done across various PRs that are wrapping up the 0.2 -> 0.3 conversion: #426, 447, #500, #501, and #507. It's likely most of those PRs will still have code changes that should be merged even after rebasing upon this work.

@retep998 At this people `RtlCopyMemory` is the only function left to be ported from the old kernel32 header, but I'm unsure how to proceed. It's a macro that either acts as a straight `memcpy` or uses a `memcpy_inline` function which wraps `memcpy` in a overlapping range check. I'm unsure how you want to handle this. If you'd rather just merge this and address that in your own commit, that might be easiest. Let me know what you want to do.

retep998:
You might want to possibly consider fixing the merge conflicts maybe.

Susurrus:
@retep998 You're planning to merge this before #447? I've been letting this sit as I figured you'd do that one first.

Susurrus:
@retep998 Ping!

PR #513, opened 2017-10-08
Migrate event APIs

skdltmxn:


retep998:
As it turns out, `eventcons` should actually be in `um`, not `shared`.

PR #524, opened 2017-11-18
Reverse order of overloaded COM methods in vtable

### LINKED ISSUE #523: Overloaded methods in COM have wrong vtable order
Apparently the COM-compatible ABI for overloaded methods is to group overloads together but put them in _reverse_ order. I observed this in my use of DirectComposition APIs, but it seems to be pretty poorly documented. The best I could find is [gcc bug 53398](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53398).

Note that Microsoft has published [guidance discouraging overloads in COM-visible interfaces](https://docs.microsoft.com/sl-si/visualstudio/code-quality/ca1402-avoid-overloads-in-com-visible-interfaces) but seems to be ignoring this more recently. I believe this renders C bindings inoperable, but that's a separate question.

Also note the previous link suggests a naming scheme of `MethodName` `MethodName_2` `MethodName_3`. I personally prefer this over the current scheme of `MN1` `MN2`, especially for methods that end in a digit, for example [`IDCompositionArithmeticCompositeEffect::SetCoefficient1`](https://msdn.microsoft.com/en-us/library/windows/desktop/dn919700.aspx). Also note that the 1-based numbering is not consistently applied, for example the overloaded methods in dwrite_3.rs use a `MN` `MN1` `MN2` convention.

I'll prepare a patch that applies the `MN` `MN_2` `MN_3` convention because I think it's best, but am open to other alternatives. It seems unlikely that we'll be breaking many existing users.

raphlinus:
Puts overloaded COM methods in reverse order to follow MSVC ABI
convention. Also changes the naming convention to use an underscore
and a 1-based number for all overrides but the first, from the
current convention which is not consistent.

Tested with animations using the DirectComposition interfaces, which
are most of the changes. The main other cluster is font management
in DirectWrite.

Fixes #523.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,5 +266,6 @@ interface IDWriteFontFileStream(IDWriteFontFileStreamVtbl): IUnknown(IUnknownVtb
     * If the duplicated thing is a struct or COM interface or union, then choose one header to be
       the canonical source of truth for that definition and publicly re-export the thing from the
       other header.
-* Sometimes a COM interface will have two methods with identical names. If the two methods are both
-  named `Foo`, then name them `Foo1` and `Foo2`.
+* Sometimes a COM interface will have two methods with identical names (an overloaded method). If
+  the two methods are both named `Foo`, then name them `Foo_1` and `Foo_2`. In addition,
+  overloaded methods must appear in *reverse* order to comply with the COM binary interface.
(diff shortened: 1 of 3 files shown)


retep998:
Could you change the style to `foo_1` `foo_2` etc and also modify `CONTRIBUTING.md` to document this requirement, especially the necessity to reverse their order.

PR #526, opened 2017-11-27
Basic WFP support

mvd-ows:
Adding definitions and constants for:
* fwptypes.h
* fwpmtypes.h
* ipsectypes.h
* iketypes.h

Adding function prototypes for:
* fwpmu.h


mvd-ows:
All the mentioned files tie into each other so there was no good good workflow for adding them one by one. Since the entire commit history is just work in progress it might make sense to squash it into a single commit.

banister:
What's the status of this pr? Looks like great work to me!

boolanger:
Not sure if this is the place to bring this up, but `FwpFreeMemory0` in _fwpmu.rs_ is potentially incorrectly defined. It's defined as a `void` function in the C header, but in this PR the definition is:
`pub fn FwpmFreeMemory0(
        p: *mut *mut c_void
    ) -> c_void;`. Calling it as it's written gives a `(exit code: 0xc000001d, STATUS_ILLEGAL_INSTRUCTION)` whereas if we remove the return type, i.e. `pub fn FwpmFreeMemory0(
        p: *mut *mut c_void
    )` there are no complaints (though I don't know if it does what it should).

retep998:
Yes, functions that return `void` in C should return nothing in Rust, **not** `c_void`.

Issue #533, opened 2017-12-18
sockaddr_in6 has disappeared from version 0.3


dimbleby:
`sockaddr_in6` used to be defined [here](https://github.com/retep998/winapi-rs/blob/77588710a683e1a4ec085b0f3d6c8939ecc7f030/src/ws2ipdef.rs#L36-L42); but I don't find anything similar in the 0.3 branch.  (The `SOCKADDR_IN` is [still defined](https://github.com/retep998/winapi-rs/blob/d8290fc406c8e390101dd225f7ec506848fc86ce/src/shared/ws2def.rs#L324-L329)).

Was this accidental or deliberate?  If accidental, any chance of a fix?

Thanks!

retep998:
This was accidental.

retep998:
Fixed by https://github.com/retep998/winapi-rs/pull/535.

PR #535, opened 2017-12-18
add SOCKADDR_IN6

dimbleby:
Per #533 

Actually, since we're breaking anyway, I've deviated from the 0.2 version which had `sockaddr_in6` and instead provided [`SOCKADDR_IN6`](https://msdn.microsoft.com/en-us/library/windows/hardware/ff570824(v=vs.85).aspx) - since this seems like it's more consistent with having `SOCKADDR_IN`.

It's possible that I haven't grokked whatever naming convention you would have used for the inner union, but so far as I can tell this definition does what I expect when I use it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -65,7 +65,7 @@ const DATA: &'static [(&'static str, &'static [&'static str], &'static [&'static
     ("winusbio", &["minwindef", "usb"], &[]),
     ("wnnc", &["minwindef"], &[]),
     ("ws2def", &["basetsd", "guiddef", "inaddr", "minwindef", "vcruntime", "winnt"], &[]),
-    ("ws2ipdef", &["in6addr", "inaddr", "minwindef"], &[]),
+    ("ws2ipdef", &["in6addr", "inaddr", "minwindef", "ws2def"], &[]),
     ("wtypes", &["guiddef", "minwindef", "ntdef", "wtypesbase"], &[]),
     ("wtypesbase", &["minwindef", "rpcndr", "winnt"], &[]),
     // ucrt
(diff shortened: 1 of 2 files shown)


retep998:
Whether the naming is consistent with other definitions is irrelevant. The only thing that matters is whether it is the same as the identifiers in the Windows SDK headers.

The naming convention for unions is documented in `CONTRIBUTING.md`.

dimbleby:
We're at cross-purposes re `sockaddr_in` vs `SOCKADDR_IN`... these are different in the Windows headers - see [here](https://msdn.microsoft.com/en-us/library/aa450948.aspx) and [there](https://msdn.microsoft.com/en-us/library/windows/hardware/ff570824(v=vs.85).aspx).  So it's not that I'm changing the name for consistency: it's that I'm changing which one is provided for consistency.

Thanks for the pointer to the naming convention - I'll make tweaks as best I understand those notes!

retep998:
MSDN is not the definitive reference on these things. According to the headers, `SOCKADDR_IN6` is a typedef to `SOCKADDR_IN6_LH` which is the actual type name. `sockaddr_in6` meanwhile is merely the struct name which by convention `winapi` does not use unless there is no type name for the struct.

dimbleby:
Well, colour me confused!

Let's not play "guess what I'm thinking"... what change, if any, would you like to see in this pull request?

retep998:
What I would like is for the PR to match the definitions in `10.0.16299.0\shared\ws2ipdef.h`.

dimbleby:
I think you mean: "please rename SOCKADDR_IN6 to SOCKADDR_IN6_LH (and make corresponding changes to the pointer type and anonymous union)".  Yes?

I'll push those changes and hope that's indeed what you intended.  If I've misunderstood, may I suggest that we'll likely draw this one to a close sooner if you would be explicit about what change you would like to see...

Thanks!

PR #504, opened 2017-09-30
Add objidl definitions

skdltmxn:

PR #544, opened 2017-12-31
Shell API additions

Jarcho:
Complete implementation for the following files:
- comcat
- idispids
- msxml
- oaidl
- objectarray
- objidl
- objsafe
- ocidl
- ocmm
- olectl
- oleidl
- propidl
- propidlbase
- shdispis
- sherrors
- shldisp
- shlguid
- shobjidl
- shobjidl_core
- shtypes
- structuredquerycondition
- urlmon
- wtypes
- xmldomdid
- xmldsodid

propidl.idl and objidl.idl both include instead of import their base counterparts, so all the definitions are duplicated. I've re-exported the base files as some definitions were mixed up between them. Alternatives would be to not re-export (backwards incompatible), or re-export only the previously misplaced definitions.

msxml has a few functions named 'parent' which conflicted with the vtable 'parent' field. Ideally, the vtable field would be renamed, but that would technically be a backwards incompatible change, so the fuction name was changed to 'parent_' instead.

### DIFF
--- a/src/um/shellapi.rs
+++ b/src/um/shellapi.rs
@@ -119,7 +119,7 @@ extern "system" {
     ) -> HICON;
     pub fn ExtractIconW(
         hInst: HINSTANCE,
-        pszExeFileName: LPCSTR,
+        pszExeFileName: LPCWSTR,
         nIconIndex: UINT,
     ) -> HICON;
 }


retep998:
*Notices PR diff is over twenty thousand lines.*

I really hate reviewing PRs sometimes...

Jarcho:
Sorry about that. Personally, I blame Microsoft.

There's a few thousand lines of overlap between this and #504, so if you wanted this one a little shorter...

PR #549, opened 2018-01-04
Added #[debug] to a small selection of structs

Lokathor:


pravic:
Came here to add `#[debug]` to the `GUID`, but it has been already done.

Is there any chance for a new winapi version to be published?

PR #559, opened 2018-01-17
Include licenses in sub-packages

Nemo157:


retep998:
No longer needed due to fa4e2c5a6dc495c71e6031bdde2e52870ebf6fdf

Nemo157:
Thanks 😄. Looking at this again I now realise that git symlinks aren't really supported on windows so this probably wouldn't work for you, even if I had noticed the explicit `include`s.

PR #564, opened 2018-02-03
Correct return values for lstrcpynW, lstrcpyW and lstrcatW functions.…

recombinant:
… These now return LPWSTR.


retep998:
Hopefully nobody was using these functions in the first place.

PR #565, opened 2018-02-05
Add IFF_* Flags

LuoZijun:
IFF Flags: `IFF_UP`, `IFF_BROADCAST`, `IFF_LOOPBACK`, `IFF_POINTTOPOINT`, `IFF_MULTICAST`

Document: https://msdn.microsoft.com/en-us/library/windows/desktop/ms738568(v=vs.85).aspx
SDK: https://github.com/LuoZijun/WindowsSDK/blob/master/10.0.16299.0/shared/ws2ipdef.h#L111-L115


retep998 (changes requested):
These flags are used for the `iiFlags` field of `INTERFACE_INFO` which has a type of `ULONG` therefore these constants should match that type.

LuoZijun:
Updated.

LuoZijun:
What can i do for that CI error ?

retep998:
Just update your branch against the latest upstream `0.3`.

LuoZijun:
Updated.

retep998 (approved):

src/shared/ws2ipdef.rs
```diff
@@ -77,3 +77,10 @@ STRUCT!{struct SOCKADDR_IN6_LH {
   u: SOCKADDR_IN6_LH_u,
 }}
 pub type PSOCKADDR_IN6_LH = *mut SOCKADDR_IN6_LH;
+
+pub const IFF_UP: ULONG = 0x00000001;
+pub const IFF_BROADCAST: ULONG = 0x00000002;
+pub const IFF_LOOPBACK: ULONG = 0x00000004;
+pub const IFF_POINTTOPOINT: ULONG = 0x00000008;
+pub const IFF_POINTOPOINT: ULONG = IFF_POINTTOPOINT;
```
retep998, line 85: As it turns out, I cannot actually find this specific definition in the Windows SDK. Where exactly did you find it?

This should likely be removed.

retep998 (changes requested):

PR #566, opened 2018-02-06
Add C type `time_t`

LuoZijun:
Add C type `time_t`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,6 +54,10 @@ pub mod ctypes {
     pub type __int64 = i64;
     pub type __uint64 = u64;
     pub type wchar_t = u16;
+    #[cfg(all(target_arch = "x86", target_env = "gnu"))]
+    pub type time_t = i32;
+    #[cfg(not(all(target_arch = "x86", target_env = "gnu")))]
+    pub type time_t = i64;
 }
 pub trait Interface {
     fn uuidof() -> shared::guiddef::GUID;


retep998:
`time_t` is not a primitive type. It is a typedef in `ucrt\corecrt.h`. It is also always a typedef to `__time64_t` (which is a typedef to `__int64`) in both the 32 and 64bit UCRT, and I have no intention of supporting a 32-bit `time_t` (what MinGW does here is irrelevant as I don't support MinGW specific stuff).

LuoZijun:
`ucrt/corecrt.h`:
Line519-Line525: https://github.com/LuoZijun/WindowsSDK/blob/master/10.0.16299.0/ucrt/corecrt.h#L519-L525

`um/IPTypes.h`:

Line30: https://github.com/LuoZijun/WindowsSDK/blob/master/10.0.16299.0/um/IPTypes.h#L30
Line99: https://github.com/LuoZijun/WindowsSDK/blob/master/10.0.16299.0/um/IPTypes.h#L99

you can see, in file `um/IPTypes.h` , they include `<time.h>` , not `um/IPTypes.h` .

so in my case, how do i translate [um/IPTypes.h:99](https://github.com/LuoZijun/WindowsSDK/blob/master/10.0.16299.0/um/IPTypes.h#L99
) ?

thx.

retep998:
Just because they include `time.h` to get the `time_t` definition, does not actually mean `time_t` was defined there. `time.h` includes other headers and somewhere along the way `ucrt\corecrt.h` gets included so it can define `time_t`.

LuoZijun:
okay, i have write like this:

```rust
pub type __time32_t = long;
pub type __time64_t = __int64;

#[cfg(target_pointer_width = "32")]
pub type time_t = __time32_t;
#[cfg(target_pointer_width = "64")]
pub type time_t = __time64_t;
```

retep998:
But `time_t` is always 64-bit, even on 32bit Windows (unless you define `_USE_32BIT_TIME_T` but `winapi` intentionally does not support that).

LuoZijun:
Oh, got it .

i will do this on `um/iptypes.rs`.

```rust
type __time64_t = __int64;
type time_t = __time64_t;
```

thank you very much.

PR #574, opened 2018-02-14
Add DwmSetWindowAttribute

SimonSapin:


SimonSapin:
This is based on squinting my eyes at `C:\Program Files (x86)\Windows Kits\10\Include\10.0.16299.0\um\dwmapi.h`. Is there a more systematic or automated way to do this?

Also, is it intended that PRs go into the `0.3` branch rather than master?

src/um/dwmapi.rs
```diff
@@ -53,7 +53,12 @@ extern "system" {
     // pub fn DwmRegisterThumbnail();
     // pub fn DwmSetDxFrameDuration();
     // pub fn DwmSetPresentParameters();
-    // pub fn DwmSetWindowAttribute();
+    pub fn DwmSetWindowAttribute(
+        hWnd: HWND,
+        dwAttribute: DWORD,
+        pvAttribute: LPVOID,
```
Arnavion: [MSDN says this is LPCVOID.](https://msdn.microsoft.com/en-us/library/windows/desktop/aa969524)

SimonSapin: My mistake. Fixed.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "winapi"
-version = "0.3.3"
+version = "0.3.4"
```
retep998: Why exactly is this commit changing this?

SimonSapin: For crates that I maintain I prefer PRs to include a version number increment, so I don’t need to make another commit before running `cargo publish`. But it sounds like that’s not your preference, I’ll revert it.

retep998: There's also the fact that the version already is 0.3.4 which makes me somewhat confused.

Arnavion: >There's also the fact that the version already is 0.3.4 which makes me somewhat confused.

Because the PR is based on a very old commit instead of latest 0.3.

https://github.com/retep998/winapi-rs/commits/3ec919bc76085922767670eb7904dcd81d736495

PR #584, opened 2018-03-16
Add dwmapi methods

raphlinus:
Add DwmFlush and DwmGetCompositionTimingInfo, both of which are useful in improving smoothness in interactive apps.


retep998:
PR does not pass CI

Issue #420, opened 2017-03-27
[Feature Request] Bindings for GDI+


snuk182:
Blocks https://github.com/snuk182/nuklear-rust/issues/1 .

docbrown:
The GDI+ structures use 8-byte packing, so I don't think this can be implemented until `#[repr(packed = "N")]` is implemented and stabilized (see rust-lang/rust#33158).

retep998:
@docbrown If the packing is greater than or equal to the struct's natural alignment, then it'll work fine. Only when the packing is lower than the struct's natural alignment, but higher than 1, that is where things are currently impossible.

PR #599, opened 2018-04-15
Gdiplus

Jarcho:
Minimal Implementation for #420.

Adds only the flat api. None of the C++ wrappers are implemented.


Connicpu:
Looks like the Self{} syntax wasn't supported yet in 1.6.0, make sure to test ^^

notgull:
Any status on this? It would be nice to be able to use this.

PR #602, opened 2018-04-16
Xaudio2

### LINKED ISSUE #593: missing bindings for XAudio2
Hi,

Any particular reason why the XAudio2 bindings are missing?

Jarcho:
Resolves #593.
Binding to xaudio2.8 instead of 2.9 due to missing import library.
All the inline functions require std for the math functions.


retep998:
What import library is missing?

Jarcho:
Ignore that part about a missing import library. Apparently xaudio2.lib links to xaudio2_9.dll.

Jarcho:
The test failure is caused by xaudio2_9.dll, which is included starting on Win 10, not existing on the test system. One of the test harnesses built on 1.6.0 links against it.

Not sure how you want to handle that error. One possibility would be to change the test script to run specific integration tests.

retep998:
The current policy for COM interfaces that do not have a UUID is to give them a UUID that is all zeros. Could you do that instead?

retep998:
Oh neat, the 1.6.0 builder is failing because it is running over the 10 minute time limit.

PR #603, opened 2018-04-16
WMI COM Interface

### LINKED ISSUE #581: Accessing hardware info (using WMI?)
I'd like to be able to access some hardware info. Normally I use WMI, but actually all I care about is reliably getting the Win32_BaseBoard's properties (e.g., SerialNumber, SKU, etc.) and the Win32_Processor's properties (e.g., ProcessorID, UniqueID, Version, etc.).

Is this possible with winapi-rs as it stands? Any links or help would be appreciated.

Jarcho:
Should resolve #581.


retep998 (approved):
Sorry for taking so long to look at this.

PR #606, opened 2018-04-22
Add event log related FFI imports

pronebird:
This PR adds missing exports for [DeregisterEventSource](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363642(v=vs.85).aspx), [RegisterEventSource](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363678(v=vs.85).aspx), [ReportEvent](
https://msdn.microsoft.com/en-us/library/windows/desktop/aa363679(v=vs.85).aspx).

### DIFF
--- a/src/um/winbase.rs
+++ b/src/um/winbase.rs
@@ -1960,20 +1960,48 @@ extern "system" {
     // pub fn BackupEventLogA();
     // pub fn BackupEventLogW();
     // pub fn CloseEventLog();
-    // pub fn DeregisterEventSource();
+    pub fn DeregisterEventSource(
+        hEventLog: HANDLE,
+    ) -> BOOL;
     // pub fn NotifyChangeEventLog();
     // pub fn GetNumberOfEventLogRecords();
     // pub fn GetOldestEventLogRecord();
     // pub fn OpenEventLogA();
     // pub fn OpenEventLogW();
-    // pub fn RegisterEventSourceA();
-    // pub fn RegisterEventSourceW();
+    pub fn RegisterEventSourceA(
+        lpUNCServerName: LPCSTR,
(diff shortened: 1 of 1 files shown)


src/um/winbase.rs
```diff
@@ -1960,20 +1960,48 @@ extern "system" {
...
-    // pub fn ReportEventW();
+    pub fn ReportEventA(
+        hEventLog: HANDLE,
+        wType: WORD,
+        wCategory: WORD,
+        dwEventID: DWORD,
+        lpUserSid: PSID,
+        wNumStrings: WORD,
+        dwDataSize: DWORD,
+        lpStrings: *const LPCSTR,
```
Connicpu: This should be `*mut LPCSTR` according to the SDK

src/um/winbase.rs
```diff
@@ -1960,20 +1960,48 @@ extern "system" {
...
+    ) -> BOOL;
+    pub fn ReportEventW(
+        hEventLog: HANDLE,
+        wType: WORD,
+        wCategory: WORD,
+        dwEventID: DWORD,
+        lpUserSid: PSID,
+        wNumStrings: WORD,
+        dwDataSize: DWORD,
+        lpStrings: *const LPCWSTR,
```
Connicpu: This should be `*mut LPCWSTR` according to the SDK

Connicpu (changes requested):

src/um/winbase.rs
pronebird: Fixed.

src/um/winbase.rs
pronebird: Fixed.

Issue #442, opened 2017-05-09
Struct size/alignment tests


retep998:
In order to ensure that the size and alignment of every struct is correct, winapi needs size and alignment tests. Currently such tests go in `structs/tests`. In order to add to this test suite, the test cases should be generated by running C++ code using VC++ with the latest Windows SDK. Every struct and union should have a test case.

```C++
// This define ensures all anonymous structs and unions have a field name so you can get its size and alignment
#define NONAMELESSUNION
// Include whichever windows SDK headers are needed to get the correct types. Note that some headers cannot be included directly.
#include <Windows.h>
#include <iostream>
#define PRINT(x) std::cout << "    assert_eq!(size_of::<"#x">(), " << sizeof(x) << ");" << std::endl;\
    std::cout << "    assert_eq!(align_of::<"#x">(), " << alignof(x) << ");" << std::endl;
int main() {
    // For regular structs and unions
    PRINT(SOME_TYPE);
    PRINT(ANOTHER_TYPE);
    // For anonymous structs and unions.
    // Unfortunately you'll have to manually rename decltype(TASKDIALOGCONFIG::u) to TASKDIALOGCONFIG_u1 in the output
    PRINT(decltype(TASKDIALOGCONFIG::u));
}
```
Copy the output into the appropriate test function, creating it if it does not exist. The order should match the order in the source file, which should match the order in the headers.

hankbao:
Why not put this in the CONTRIBUTING? I can make a PR if you like.

retep998:
What I actually want is something that can automatically generate these tests...

hankbao:
Fine. But I end up rolling my own `PRINT_TEST` before I dug this issue out. I bet other contributors would be more likely to write tests in their PRs if they saw this code snippet in the CONTRIBUTING.

Coding-Badly:
I'm starting a translation of Windows Filtering Platform.  There are multiple unnamed unions.  As far as I can tell `#define NONAMELESSUNION` has no effect.  Using `decltype` is mentioned in `CONTRIBUTING.md`.  How do I use `decltype` to name an anonymous union?

retep998:
@Coding-Badly The naming of anonymous unions is covered under https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#anonymous-unions-and-structs

Coding-Badly:
[...]

PR #607, opened 2018-04-22
Add full set of struct tests + changes to make them pass

Jarcho:
Issue #442.

Fix summary:
- ntddscsi had some of the structs put behind an IFDEF to match the sdk header.
- Quite a few zero sized arrays were changed to one to match the sdk headers. They weren't really wrong, but the tests on them fail otherwise.
- Quite a few struct were not packed when they should have been.
- A few structs were changed to contain pointers instead of the pointee type.
- A couple of BOOL and BOOLEAN mixups.
- A few structs were changed to unions.
- Some unions had the wrong storage type.
- Some missing fields were added.

One unresolved issue is SPPHRASE in sapi. The C and C++ definitions differ from each other and I'm unsure as to which is the correct definition. The C++ definition uses base classes which add padding the C definition does not.

Would you like the C code used to generate the tests added in as well?


src/um/commdlg.rs
```diff
@@ -526,13 +526,13 @@ interface IPrintDialogServices(IPrintDialogServicesVtbl): IUnknown(IUnknownVtbl)
...
 );
-STRUCT!{struct PRINTPAGERANGE {
+STRUCT!{#[cfg_attr(target_arch = "x86", repr(packed))]  struct PRINTPAGERANGE {
     nFromPage: DWORD,
     nToPage: DWORD,
 }}
 pub type LPPRINTPAGERANGE = *mut PRINTPAGERANGE;
 pub type PCPRINTPAGERANGE = *const PRINTPAGERANGE;
-STRUCT!{struct PRINTDLGEXA {
+STRUCT!{#[cfg_attr(target_arch = "x86", repr(packed))]  struct PRINTDLGEXA {
```
retep998: Are these... double spaces I see? You monster!

src/um/setupapi.rs
```diff
@@ -32,25 +32,25 @@ pub const MAX_SERVICE_NAME_LEN: usize = 256;
 pub const MAX_SUBTITLE_LEN: usize = 256;
 pub const SP_MAX_MACHINENAME_LENGTH: usize = MAX_PATH + 3;
 pub type HINF = PVOID;
-STRUCT!{struct INFCONTEXT {
+STRUCT!{#[cfg_attr(target_arch = "x86", repr(packed))]  struct INFCONTEXT {
```
retep998: And even more of those double spaces!

retep998 (changes requested):
Thanks for putting all the work into this PR. Just watch out for those spaces.

retep998 (approved):

PR #611, opened 2018-04-26
Add endpointvolume

MSxDOS:


sbstp:
There's already an audioclient module with APIs related to core audio, does this need its own module?

retep998:
@sbstp  It's not about whether something needs its own module. It's about everything should always be in the module that corresponds to the header it is from. Since these definitions are from `um\endpointvolume.h` they belong in the `um::endpointvolume` module.

sbstp:
I rewrote my sample program with this API instead, it turns out that this is much simpler to use. And it actually works! Thanks @MSxDOS ! This program toggles mute/unmute for the OS.
```rust
extern crate winapi;

mod com;

use std::ptr;

use winapi::shared::wtypesbase::*;
use winapi::Interface;
use winapi::shared::ntdef::NULL;
use winapi::um::combaseapi::*;
use winapi::um::mmdeviceapi::*;
use winapi::shared::minwindef::*;
use winapi::um::endpointvolume::*;

use com::ComPtr;

fn main() {
    unsafe {
        CoInitializeEx(NULL, COINITBASE_MULTITHREADED);
        
        let mut enumerator: *mut IMMDeviceEnumerator = ptr::null_mut(); 
        assert!(CoCreateInstance(&CLSID_MMDeviceEnumerator, std::ptr::null_mut(), CLSCTX_INPROC_SERVER, &IMMDeviceEnumerator::uuidof(), &mut enumerator as *mut _ as *mut _) == 0);
        let enumerator = ComPtr::from_raw(enumerator);

        let mut endpoint: *mut IMMDevice = ptr::null_mut();
        assert!(enumerator.GetDefaultAudioEndpoint(eRender, eMultimedia, &mut endpoint as *mut _) == 0);
        let endpoint = ComPtr::from_raw(endpoint);

        let mut volume: *mut IAudioEndpointVolume = ptr::null_mut();
        assert!(endpoint.Activate(&IAudioEndpointVolume::uuidof(), CLSCTX_ALL, ptr::null_mut(), &mut volume as *mut _ as *mut _) == 0);
        let volume = ComPtr::from_raw(volume);

        let mut muted: BOOL = 0;
        volume.GetMute(&mut muted as *mut _);
        let muted = muted != 0;
        volume.SetMute((!muted) as BOOL, ptr::null_mut());

        CoUninitialize();
    }
}
```

PR #621, opened 2018-06-05
bindings for dbgeng.h and wdbgexts.h

### LINKED ISSUE #619: Bindings for wdbgext.h and dbgeng.h
I need bindings to work with the Windows Debugger.

I know I'll need parts of wdbgext.h and dbgeng.h to start with.

I'll submit a PR if an incomplete translation is acceptable. dbgeng.h is 25000+ lines (though quite a bit of duplication) and wdbgext.h has nearly 50 inline functions that I'm not keen on translating.

lzybkr:
I made a significant shortcut that works correctly, but isn't a literal translation.

dbgeng.h duplicates many method definitions instead of using inheritance, presumably to enable use from C.

In my conversion to Rust, I used inheritance to avoid the duplication.

Fixes #619.


lzybkr:
The failure with `TARGET=1.6.0-x86_64-pc-windows-gnu` is from the use of `...`, but it works everywhere else.

I don't know the history of variadic FFI functions - but they seem to have been around for a long time - before v1.0 at least - so why would it fail with just this one config?

retep998:
It's not the variadic FFI function, but rather the macro itself. Macros have had a lot of subtle changes over the versions, so it's easy to run into failure cases.

garyttierney:
Hi guys, sorry for the ping. I can see there's quite the monumental backlog of open PRs!

Is there anything that needs done to move this forward on the dev side? I've rebased the branch on the latest changes recently and would be interested in getting the PR updated.

lzybkr:
This is a huge PR so I don't blame @retep998 for not finding time to review it. I mostly opened the PR as a way of sharing, not fully expecting it to be merged in any reasonable time frame.

I do appreciate knowing PRs aren't merged without a thorough review as it's hard to test changes accepted here and errors introduced in this crate can easily cause security issues due to unsafe code.

I'm not sure if @retep998 is open to adding additional maintainers - I imagine I could find some willing Microsoft employees to pitch in if that was an option. Or maybe @retep998 is fine with accepting sign off from folks on specific PRs like this on a case by case basis?

@garyttierney - I could rebase my branch - but I'm wondering if it makes sense to move this PR to it's own repo and add nice Rust bindings to make it easier to work with the Windows debugger apis. I have some wrappers similar to what you have [here](https://github.com/garyttierney/dbgadapter/tree/master/backends/windbg/src/dbgeng).

So rough proposal - I create a new repo called `dbgeng-rs` which has 2 crates - `dbgeng-rs` and `dbgeng-sys`, this PR moves to `dbgeng-sys`, and we collaborate on `dbgeng-rs`? I could see a project like [headcrab](https://github.com/headcrab-rs/headcrab) taking advantage of this as well.

garyttierney:
Hi @lzybkr, just caught your comment.

> So rough proposal - I create a new repo called dbgeng-rs which has 2 crates - dbgeng-rs and dbgeng-sys, this PR moves to dbgeng-sys, and we collaborate on dbgeng-rs? I could see a project like headcrab taking advantage of this as well.

Yep, this sounds like a good idea to me. I'm happy to collaborate on this and get some nice bindings developed. Also, thanks for bringing headcrab to my attention!

I imagine the `-sys` bindings could be later adopted back into this project, following the embrace-extend-extinguish approach if that's something retep998 is interested in.

PR #623, opened 2018-06-07
Added missing licenses for subcrates for compliance reasons.

aatxe:
This is a version of #559 that actually includes the licenses in the subcrates which (though I am not a lawyer and this does not constitute legal advice) I believe to be required for license compliance reasons.


retep998:
This PR is missing necessary changes to the `Cargo.toml` files, so I did fa4e2c5a6dc495c71e6031bdde2e52870ebf6fdf instead.

Issue #647, opened 2018-07-26
Add D3DCOLOR_x macros


abhijeetbhagat:
I propose adding the following color conversion macros:

```
D3DCOLOR_ARGB(a,r,g,b)
D3DCOLOR_RGBA(r,g,b,a)
D3DCOLOR_XRGB(r,g,b)
D3DCOLOR_XYUV(y,u,v)
D3DCOLOR_AYUV(a,y,u,v)
D3DCOLOR_COLORVALUE(r,g,b,a)
```

Can I add them?

retep998:
Well, they would be added as functions, not macros, but yes.

abhijeetbhagat:
Why not macros? I meant something like:
```
macro_rules! D3DCOLOR_ARGB {
    ($a:expr, $r:expr, $g:expr, $b:expr) => {
       ((($a & 0xff) << 24) | (($r & 0xff) << 16) | (($g & 0xff) << 8) | ($b & 0xff)) as D3DCOLOR
    }
}

macro_rules! D3DCOLOR_RGBA {
    ($r:expr, $g:expr, $b:expr, $a:expr) => {
        D3DCOLOR_ARGB!($a, $r, $g, $b)
    }
}

macro_rules! D3DCOLOR_XRGB {
    ($r:expr, $g:expr, $b:expr) => {
        D3DCOLOR_ARGB!(0xff, $r, $g, $b)
    }
}

macro_rules! D3DCOLOR_XYUV {
    ($y:expr, $u:expr, $v:expr) => {
        D3DCOLOR_ARGB!(0xff, $y, $u, $v)
    }
}

macro_rules! D3DCOLOR_AYUV {
    ($a:expr, $y:expr, $u:expr, $v:expr) => {
        D3DCOLOR_ARGB!($a, $y, $u, $v)
    }
}

macro_rules! D3DCOLOR_COLORVALUE {
    ($r:expr, $g:expr, $b:expr, $a:expr) => {
        D3DCOLOR_ARGB!(($r * 255f32) as DWORD, ($g * 255f32) as DWORD, ($b *255f32) as DWORD, (($a *255f32) as DWORD))
    }
}
```

retep998:
These macros are literally invoked as if they are functions, so they should be implemented as inline functions. Rust macros are only used when strictly necessary for constants.

https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#inline-functions-and-macros

[...]

PR #649, opened 2018-07-27
Add DX related color conversion functions

abhijeetbhagat:
Add D3DCOLOR_x macros #647


retep998:
Can you get rid of the blank lines and move the function definitions between the definition for `D3DCOLOR` and `D3DVECTOR`? Thanks.

PR #655, opened 2018-08-01
FIELD_OFFSET

MSxDOS:
@retep998 Do you want this one?


retep998:
As useful as this macro is for Windows FFI, it is useless for writing FFI bindings, and can be trivially provided by a crate independent of `winapi` where it would be useful for *all* FFI in general, and not just Windows FFI.

PR #666, opened 2018-08-21
Make the inner field of a tuple struct defined by UNION! public

yying:
Currently, the inner field of the defined tuple structure is private for `UNION!`s. This creates issues in situations where we want to modify fields within `UNION!` defined tuple structs. Making the inner field public alleviates this problem.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -253,7 +253,7 @@ macro_rules! UNION {
         $($variant:ident $variant_mut:ident: $ftype:ty,)+
     }) => (
         #[repr(C)] $(#[$attrs])*
-        pub struct $name([$stype; $ssize]);
+        pub struct $name(pub [$stype; $ssize]);
         impl Copy for $name {}
         impl Clone for $name {
             #[inline]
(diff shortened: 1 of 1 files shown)


retep998:
In what situation do you need to modify the inner field? The representation is intentionally private so that I can switch to real unions in a future version without breaking code that chose to rely on internal details of the current implementation.

yying:
I'm trying to do the following:
<pre>
let mut p = ... as PSAMPLE_STRUCT;
(*p).u1 = SAMPLE_STRUCT_u1([...]);
</pre>
where the `u1` field is a `UNION!` defined "union".

I want to set `u1` to a new value. In the current form, this code will result in an error stating that the "constructor is not visible here due to private fields". 

Is there a better way to accomplish this?

retep998:
User the setter methods on `u1` to assign a value to the specific union variant.

yying:
Sounds good, are the setters already implemented for items defined by the `UNION!` macro? I only see getters as per https://github.com/retep998/winapi-rs/blob/661685f97d7aa03e1a5832b022bd2c08ea171e49/src/macros.rs#L267-L276 and https://github.com/retep998/winapi-rs/blob/661685f97d7aa03e1a5832b022bd2c08ea171e49/src/macros.rs#L296-L305

retep998:
The `variant_mut` methods give you a `&mut T` which allows you to mutate that variant of the union.

yying:
Got it. Thanks for the help! Going to close this PR as it is not relevant anymore.

PR #672, opened 2018-08-31
Add pub modifier to tuple constructor in UNION macro

mosyp:
Generated structs by UNION macros has no pub modifier in its tuple constructor, hence their creation is not possible:
```
50 |         let u = OVERLAPPED_u([0]);
   |                 ^^^^^^^^^^^^ constructor is not visible here due to private fields
```
Macro produces:
```rust
pub struct OVERLAPPED_u([u64; 1]); // before
pub struct OVERLAPPED_u(pub [u64; 1]); // after
```

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -253,7 +253,7 @@ macro_rules! UNION {
         $($variant:ident $variant_mut:ident: $ftype:ty,)+
     }) => (
         #[repr(C)] $(#[$attrs])*
-        pub struct $name([$stype; $ssize]);
+        pub struct $name(pub [$stype; $ssize]);
         impl Copy for $name {}
         impl Clone for $name {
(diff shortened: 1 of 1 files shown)


retep998:
The internal representation of unions is *intentionally* private so that people cannot rely on that implementation detail, allowing `winapi` to upgrade to using real unions in a future version without breaking any code. The way you're supposed to construct instances of unions until then is `std::mem::zeroed()`. It's even in the [README](https://github.com/retep998/winapi-rs#how-do-i-create-an-instance-of-a-union)!

Also this is a duplicate of https://github.com/retep998/winapi-rs/pull/666

alkeryn:
@retep998 hey i know it's an old question, but i saw that we can use Default instead of zeroed, should we ?

PR #680, opened 2018-09-18
add newer DWrite interfaces necessary for working with variation fonts

lsalzman:
These interfaces are necessary so we can extend WebRender in Mozilla Firefox to support variation fonts on Windows (see https://bugzilla.mozilla.org/show_bug.cgi?id=1441323)


lsalzman:
Thank you

PR #684, opened 2018-09-20
Add missing parts of dsound

hiddenhare:
Brings dsound.rs up to par with dsound.h.

Some quirks:
- The I3DL2 environment presets have been translated into const `DSFXI3DL2Reverb` structs
- The I3DL2 material presets have not been implemented, since they seem to be an undocumented addition and there's no obvious way to use them with the rest of the API
- `DirectSoundFullDuplexCreate8` (in the original header, a `#define` for `DirectSoundFullDuplexCreate`) has been implemented as an inline function


retep998:
PR has conflicts.

hiddenhare:
Conflicts resolved.

PR #687, opened 2018-09-27
Add Native Wifi / WLAN API support

### LINKED ISSUE #624: Native Wifi / Wlanapi
[MSDN Native Wifi Reference](https://msdn.microsoft.com/en-us/library/windows/desktop/ms706275(v=vs.85).aspx) -> `Wlanapi.h` etc.

Is this missing and wanted?

MSxDOS:
## Notes
- Some constants may need to be assigned proper winapi'sh types, like all the OID_DOT11_... in `windot11` which are probably `ULONG` but I didn't want to make any assumptions.
- `ntddndis` is _a little_ incomplete. I only added what was required and left it as-is since it doesn't seem to be a part of WLAN API anyway.
- Ping me if something that should be there is missing.  

Otherwise, Closes #624 :tada:


src/shared/windot11.rs
```diff
@@ -0,0 +1,2854 @@
...
+pub const DOT11_MAX_NUM_DEFAULT_KEY: i32 = 4;
+pub const DOT11_MAX_NUM_DEFAULT_KEY_MFP: i32 = DOT11_MAX_NUM_DEFAULT_KEY + 2;
+pub const OID_DOT11_NDIS_START: u32 = 0x0D010300;
+pub const NWF_MANDATORY_OID: u32 = 0x01;
+pub const NWF_OPTIONAL_OID: u32 = 0x02;
+pub const NWF_OPERATIONAL_OID: u32 = 0x01;
+pub const NWF_STATISTICS_OID: u32 = 0x02;
+#[inline]
+pub fn NWF_DEFINE_OID(Seq: u32, o: u32, m: u32) -> u32 {
+    0x0E000000 | o << 16 | m << 8 | Seq
```
retep998: Can there be some parentheses here to make the order unambiguous?

retep998 (changes requested):

PR #699, opened 2018-10-20
Add ConPty APIs

### LINKED ISSUE #695: Add support for the new ConPTY API
The API allows for communicating with CLI processes without passing through cmd.

~~It's currently only available on insider builds~~ and is only documented (as far as I can tell) in [this blog post](https://blogs.msdn.microsoft.com/commandline/2018/08/02/windows-command-line-introducing-the-windows-pseudo-console-conpty/).

I'm not sure what the policies for dealing with insider APIs is, but this would be very useful for the implementation of the API in alacritty. See the [downstream issue](https://github.com/zacps/alacritty/issues/25).

davidhewitt:
Closes #695.

I added the bare minimum to cover those three functions. These APIs will only work on the very newest Windows. Does anything need to be done to feature gate them?


retep998:
There's no need to feature gate things based on Windows version because extern functions are only linked to when used. That said, if you *do* use these functions then your program won't run at all on older Windows versions unless you use `GetProcAddress` to obtain pointers to the functions instead of using these extern functions directly.

davidhewitt:
Cool. I'd rather not use `GetProcAddress` but it's going to be quite a small segment of windows users able to use these currently! 😄 

Do you need anything else to make this PR mergeable?

chyyran:
1809 is now GA, so ConPTY should be available on updated Windows 10 systems.

src/um/wincon.rs
```diff
@@ -538,3 +538,4 @@ extern "system" {
         ExeNameBufferLength: DWORD,
     ) -> DWORD;
 }
+pub type HPCON = *mut c_void;
```
retep998: This is actually defined in `wincontypes.h`

retep998 (changes requested):

src/um/wincon.rs
davidhewitt: I can move it to a new file "wincontypes.rs". Do I need to add an accompanying feature for this new file?

I'd originally not done so because it seemed a bit overkill to make a new file for just the one line! While I'm at it want me to attempt to wrap whatever else is defined in `wincontypes.h`?

retep998: You would need to add a feature for the new file. It's up to you whether you want to bind whatever else is in there.

davidhewitt: Great - will see how much I can find time for this weekend.

davidhewitt:
I've moved HPCON to `wincontypes.rs` and introduced a new feature as discussed. Turns out a number of the struct definitions in `wincon.rs` are also defined in `wincontypes.h`, so I've moved those at the same time.

retep998:
When moving existing definitions to a different header, please put in `pub use` statements so that code still relying on the definitions in the old location is not broken.

Also when adding a new header, you also have to update the feature array in `build.rs`.

davidhewitt:
Thanks - will fixup tomorrow morning :)

davidhewitt:
Have corrected all that now. Many thanks for your patience in helping refine all the bits I originally missed. Would a separate PR be useful to extend the notes in CONTRIBUTING to make those points clearer?

davidhewitt:
Just wondering if there's anything else you'd like to see to have this merged?

davidhewitt:
@retep998 just pinging to say we're hoping to merge https://github.com/jwilm/alacritty/pull/1762 but would like to see these definitions merged into winapi first (we're using this patch I'm providing). Please let me know if there's anything left I can do to help this PR merge.

retep998 (approved):

davidhewitt:
Thanks!

PR #702, opened 2018-10-23
Add interface identifier conversion

faern:
Adding bindings to functions allowing converting from network interface aliases, names, GUIDs and indexes to their `NET_LUID` representation, and back again.


src/shared/netioapi.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright © 2015-2017 winapi-rs developers
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// All files in the project carrying such notice may not be copied, modified, or distributed
+// except according to those terms.
+use shared::basetsd::SIZE_T;
+use shared::guiddef::GUID;
+use shared::ifdef::{NET_IFINDEX,NET_LUID,PNET_IFINDEX,PNET_LUID};
```
retep998: There should be a space after the comma.

src/shared/netioapi.rs
```diff
@@ -0,0 +1,49 @@
...
+use shared::minwindef::DWORD;
+use shared::ntdef::{PWSTR,WCHAR};
```
retep998: Here as well.

src/shared/ifdef.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright © 2015-2017 winapi-rs developers
```
retep998: Shouldn't the copyright year be just 2018?

src/shared/netioapi.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright © 2015-2017 winapi-rs developers
```
retep998: Here as well.

retep998 (changes requested):

src/shared/netioapi.rs
faern: Fixed

src/shared/ifdef.rs
faern: Fixed

src/shared/netioapi.rs
faern: Fixed

src/shared/netioapi.rs
faern: Fixed

faern:
I added the `A` version where there were `W` versions present. Also squashed everything to one commit and added the PR number to the commit message, as seems to be the standard here.

retep998:
The adding of the PR number to the commit message is done by the Squash and merge feature of Github which I use to merge PRs.

faern:
All feedback has been addressed. So I think this is ready for merge?

retep998 (approved):

PR #710, opened 2018-11-19
Add msdelta

topecongiro:
Hi, I am interested in using Delta Compression library from Rust and trying to add msdelta to this library. So far I have added what seems to be the minimal interfaces needed to experiment with the APIs. However, when I try to build the following code, I get a link error that says it cannot find `CreateDeltaA`:

```rust
use winapi::um::msdelta::*;
use winapi::um::sysinfoapi::*;

pub fn diff() {
    let mut delta_input_u: DELTA_INPUT_U = unsafe { std::mem::zeroed() };
    unsafe {
        delta_input_u.lpcStart_mut();
    }
    let delta_input = DELTA_INPUT {
        uSize: 0,
        Editable: false,
        u: delta_input_u,
    };
    let mut system_time = unsafe { std::mem::zeroed() };
    unsafe {
        GetSystemTime(&mut system_time);
    }
    let mut file_time = unsafe { std::mem::zeroed() };
    unsafe {
        GetSystemTimeAsFileTime(&mut file_time);
    }
    let lpSourceName = 0;
    let lpTargetName = 0;
    let lpSourceOptionsName = 0;
    let lpTargetOptionsName = 0;
    let lpDeltaName = 0;
    unsafe {
        let _result = CreateDeltaA(
            1,
            0,
            0,
            &lpSourceName,
            &lpTargetName,
            &lpSourceOptionsName,
            &lpTargetOptionsName,
            delta_input,
            &file_time,
            0,
            &lpDeltaName,
        );
    }
}
```

Any idea on what I am doing wrong and what I should do from here? I am fairly new to working on Rust in Windows, so my apologies if I am missing something very basic.


build.rs
```diff
@@ -221,6 +221,7 @@ const DATA: &'static [(&'static str, &'static [&'static str], &'static [&'static
     ("msaatext", &[], &[]),
     ("mscat", &["guiddef", "minwindef", "mssip", "wincrypt", "winnt"], &[]),
     ("mschapp", &["basetsd", "minwindef", "winnt"], &["advapi32"]),
+    ("msdelta", &["basetsd", "minwindef", "wincrypt", "winnt"], &[]),
```
retep998: You have to add `"msdelta"` to the third field here since that is the name of the import library that contains the symbols you are trying to link to.

topecongiro: Thanks, it worked 🎉

retep998:
```
memory allocation of 738197504 bytes failed
```
Uhhhh. This doesn't sound good.

src/um/msdelta.rs
```diff
@@ -0,0 +1,158 @@
...
+// All files in the project carrying such notice may not be copied, modified, or distributed
+// except according to those terms.
+//! Delta Compression Engine API.
+use ctypes::__int64;
+use shared::basetsd::SIZE_T;
+use shared::minwindef::{BOOL, DWORD, FILETIME, LPCVOID, LPVOID, UCHAR};
+use um::wincrypt::ALG_ID;
+use um::winnt::{LPCSTR, LPCWSTR};
+UNION!{union DELTA_INPUT_U {
+    [u64; 1],
```
topecongiro: This ought to be `[u32; 1]` 😞

topecongiro:
Hopefully https://github.com/retep998/winapi-rs/pull/710/commits/e10e61919c60fdd7515c489d78e33a9d5fc69722 would fix the build error...

src/um/msdelta.rs
retep998: Actually it ought to be `[usize; 1]` as the union variants are all pointers.

topecongiro: Hmm, changing it to `[usize; 1]` does not work locally: `CreateDeltaA` fails with error code 13 (`ERROR_INVALID_DATA`).

topecongiro: Ohhh, I just realized that I was mixing up the order of fields of `DELTA_INPUT`. That is why I was getting 13 error code when using `[usize; 1]`. Sorry for missing such an obvious mistake!

https://github.com/retep998/winapi-rs/pull/710/commits/93ac610505732e2fbfe17802b33a0c4afcd20073 should fix the issue.

topecongiro:
@retep998 Sorry for the delay, I have updated the PR, could you review this please?

retep998:
While I'm sure you put a lot of effort into the doc comments, I'd rather not have them in `winapi`. Given that inline documentation will never be as good as the comprehensive documentation on MSDN, minimizing crate size takes priority.

topecongiro:
> While I'm sure you put a lot of effort into the doc comments, I'd rather not have them in winapi. Given that inline documentation will never be as good as the comprehensive documentation on MSDN, minimizing crate size takes priority.

I see your point. Removed them all in https://github.com/retep998/winapi-rs/pull/710/commits/68b0d9fc0df9ffd69f65e4a3220b1d8237963329.

src/um/msdelta.rs
```diff
@@ -0,0 +1,283 @@
...
+pub const DELTA_FILE_TYPE_AMD64: DELTA_FILE_TYPE = 0x00000008;
+pub const DELTA_FILE_TYPE_CLI4_I386: DELTA_FILE_TYPE = 0x00000010;
+pub const DELTA_FILE_TYPE_CLI4_AMD64: DELTA_FILE_TYPE = 0x00000020;
+pub const DELTA_FILE_TYPE_CLI4_ARM: DELTA_FILE_TYPE = 0x00000040;
+pub const DELTA_FILE_TYPE_CLI4_ARM64: DELTA_FILE_TYPE = 0x00000080;
+pub const DELTA_FILE_TYPE_SET_RAW_ONLY: DELTA_FILE_TYPE = DELTA_FILE_TYPE_RAW;
+pub const DELTA_FILE_TYPE_SET_EXECUTABLES_1: DELTA_FILE_TYPE = DELTA_FILE_TYPE_RAW
+                                                             | DELTA_FILE_TYPE_I386
+                                                             | DELTA_FILE_TYPE_IA64
+                                                             | DELTA_FILE_TYPE_AMD64;
```
retep998: Do not use aligned indentation.

retep998 (changes requested):

src/um/msdelta.rs
topecongiro: https://github.com/retep998/winapi-rs/pull/710/commits/8ac2ea3eecbffcb3021a8b2a7351ddacb3d73040 Fixed it.

.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 Cargo.lock
+.idea
```
retep998, line 3: As this folder is specific to your choice of IDE, it should be added to your own local git configuration's `.gitignore`, not the one here.

retep998 (changes requested):

G33KatWork:
Is there anything else than the requested removal of the `.idea` entry in `.gitignore` preventing this from being merged? I could really use this binding right now.

Issue #728, opened 2019-01-15
winapi fails extended cargo test on link.exe with aarch64-pc-windows-msvc


EwoutH:
**AppVeyor config:**
```
image: Visual Studio 2017
install:
  - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
  - rustup-init -yv --default-toolchain nightly
  - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%
  - ps: $Env:PKG_CONFIG_ALLOW_CROSS=1
  - rustc -vV
  - cargo -vV
  - rustup target add aarch64-pc-windows-msvc
build_script:
  - cargo build --target=aarch64-pc-windows-msvc
  - cargo build --target=aarch64-pc-windows-msvc --features "everything debug impl-default"
test_script:
  - cargo test --target=aarch64-pc-windows-msvc --features "everything debug impl-default"
```
Both builds run without problems, but the test gives an error on link.exe. It could be an external problem.

**Crash log:**
```
cargo test --target=%TARGET% --features "everything debug impl-default"
   Compiling winapi v0.3.6 (C:\projects\winapi-rs)
error: linking with `link.exe` failed: exit code: 1
  |
[...]

Issue #727, opened 2019-01-15
rav1e: winapi does not compile to aarch64-pc-windows-msvc target


EwoutH:
When running `cargo build --release --target=aarch64-pc-windows-msvc` winapi-rs doesn't compile. Is there something I need to update or configure differently, or is this a bug?

Full log: https://ci.appveyor.com/project/EwoutH/rav1e/builds/21624989/job/anr30v88gfie2cx5

```
error[E0432]: unresolved import `um::winnt::PIMAGE_NT_HEADERS`
  --> C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\winapi-0.3.6\src\um\dbghelp.rs:14:55
   |
14 |     BOOLEAN, CHAR, HANDLE, LIST_ENTRY, PCSTR, PCWSTR, PIMAGE_NT_HEADERS, PIMAGE_SECTION_HEADER,
   |                                                       ^^^^^^^^^^^^^^^^^ no `PIMAGE_NT_HEADERS` in `um::winnt`. Did you mean to use `PIMAGE_NT_HEADERS64`?
error[E0432]: unresolved import `um::winnt::PCONTEXT`
  --> C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\winapi-0.3.6\src\um\errhandlingapi.rs:11:48
   |
11 |     EXCEPTION_POINTERS, LONG, LPCSTR, LPCWSTR, PCONTEXT, PEXCEPTION_RECORD,
   |                                                ^^^^^^^^ no `PCONTEXT` in `um::winnt`
error[E0432]: unresolved import `um::winnt::PCONTEXT`
  --> C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\winapi-0.3.6\src\um\minwinbase.rs:22:52
   |
22 |     CHAR, EXCEPTION_RECORD, HANDLE, LPSTR, LPWSTR, PCONTEXT, PRTL_CRITICAL_SECTION,
   |                                                    ^^^^^^^^ no `PCONTEXT` in `um::winnt`
error[E0432]: unresolved import `um::winnt::CONTEXT`
  --> C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\winapi-0.3.6\src\um\processthreadsapi.rs:16:5
   |
[...]

PR #729, opened 2019-01-15
Reconfigure appveyor.yml to enable aarch64-pc-windows-msvc target

EwoutH:
This PR reconfigures the appveyor.yml to enable building and testing to `aarch64-pc-windows-msvc` targets. It changes the following:

- Sets default `image` to `Visual Studio 2017`
- Builds Rust with `rustup-init` instead of the different `rust-install.exe`
- Allows cross compiling with `$Env:PKG_CONFIG_ALLOW_CROSS=1``
- Adds `aarch64-pc-windows-msvc` target

It should help with future testing of `aarch64-pc-windows-msvc` buids and issue #727. Building on `aarch64-pc-windows-msvc` works, even with extended features, but when tested with cargo `link.exe` throws an error (see #728).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,7 @@
 environment:
   matrix:
   - TARGET: nightly-x86_64-pc-windows-msvc
+  - TARGET: nightly-aarch64-pc-windows-msvc
   - TARGET: nightly-i686-pc-windows-msvc
   - TARGET: nightly-x86_64-pc-windows-gnu
   - TARGET: nightly-i686-pc-windows-gnu
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,6 @@
 environment:
+  channel: nightly
+  host: x86_64-pc-windows-msvc
   matrix:
   - TARGET: nightly-x86_64-pc-windows-msvc
   - TARGET: nightly-aarch64-pc-windows-msvc
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,9 +25,8 @@ matrix:
     - target: aarch64-pc-windows-msvc
     
 cache:
- - C:\Users\appveyor\.rustup
- - C:\Users\appveyor\.rustc
- - C:\Users\appveyor\.cargo
+ - C:\projects\winapi-rs
+ - C:\Users\appveyor
 
 install:
   - ps: if (ls -r . -fi "*.rs" | sls "`t") { throw "Found tab character" }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,8 +19,7 @@ environment:
     channel: stable
   - target: x86_64-pc-windows-gnu
     host: x86_64-pc-windows-gnu
-    channel: none
-    version: 1.6.0
+    channel: 1.6.0
 matrix:
   allow_failures:
     - target: aarch64-pc-windows-msvc
@@ -29,7 +28,6 @@ install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
   - rustup-init -yv --default-toolchain %channel% --default-host %host%
   - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%
-  - rustup install %version%
   - rustup target add %target%
   - rustc -vV
   - cargo -vV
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,22 +1,35 @@
+image: Visual Studio 2017
+
 environment:
+  channel: nightly
   matrix:
-  - TARGET: nightly-x86_64-pc-windows-msvc
-  - TARGET: nightly-i686-pc-windows-msvc
-  - TARGET: nightly-x86_64-pc-windows-gnu
-  - TARGET: nightly-i686-pc-windows-gnu
-  - TARGET: 1.6.0-x86_64-pc-windows-gnu
```
retep998, line 7: Please preserve testing of 1.6.0 as it is important for not making breaking changes to the minimum supported Rust version.

appveyor.yml
```diff
@@ -1,22 +1,35 @@
...
+
 install:
-  - ps: if (ls -r . -fi "*.rs" | sls "`t") { throw "Found tab character" }
-  - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-${env:TARGET}.exe" -FileName "rust-install.exe"
-  - ps: .\rust-install.exe /VERYSILENT /NORESTART /DIR="C:\rust" | Out-Null
-  - ps: $env:PATH="$env:PATH;C:\rust\bin"
+  - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
+  - rustup-init -yv --default-toolchain %channel% --default-host %HOST%
+  - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%
+  - ps: $Env:PKG_CONFIG_ALLOW_CROSS=1
```
retep998: What is this environment variable even used for? There's no pkgconfig dependencies anywhere so why would this ever apply?

retep998 (changes requested):

appveyor.yml
EwoutH: You're right, I removed it

appveyor.yml
EwoutH: Added the 1.6.0 target back

appveyor.yml
```diff
@@ -1,22 +1,44 @@
...
+    channel: stable
+  - target: x86_64-pc-windows-gnu
+    host: x86_64-pc-windows-gnu
+    channel: 1.6.0
+matrix:
+  allow_failures:
+    - target: aarch64-pc-windows-msvc
+
 install:
-  - ps: if (ls -r . -fi "*.rs" | sls "`t") { throw "Found tab character" }
```
retep998: This check needs to be retained.

appveyor.yml
```diff
@@ -1,22 +1,44 @@
...
-  - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-${env:TARGET}.exe" -FileName "rust-install.exe"
-  - ps: .\rust-install.exe /VERYSILENT /NORESTART /DIR="C:\rust" | Out-Null
-  - ps: $env:PATH="$env:PATH;C:\rust\bin"
+  - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
+  - rustup-init -yv --default-toolchain %channel% --default-host %host%
+  - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%
```
retep998: What is `%APPVEYOR_BUILD_FOLDER%` for?

retep998 (changes requested):

appveyor.yml
EwoutH: Good spot! It's for easily locating where AppVeyor stored the builds produced so that they can be published as an artifact (enabled in latest commit).

retep998: But why do we need artifacts published?

EwoutH: No particular reason, was experimenting a bit. In this case artifacts are not that useful indeed, I removed them.

EwoutH:
Thanks for the patience and critical feedback, opened a PR a little too soon. Should be ready to merge now!

retep998:
Thanks for putting in the work to get aarch64 tested. I'll see if I can't get CI tests passing there shortly.

EwoutH:
Thanks, if you need any help let me know!

PR #735, opened 2019-01-28
Let `#define`d functions on 32-bit have same signatures as on 64-bit.

IslandUsurper:
Because Rust treats isize and i32 as different types even when they are the same size, aliasing `SetWindowLongPtrW` to `SetWindowLongW` doesn't work without a cast. So, let's do the cast in winapi on x86.


retep998:
This is a breaking change that will have to be postponed until the next major version.

IslandUsurper:
@retep998, should I redo this PR against master, then?

retep998:
No, because that is not what `master` is for (and also I can change what branch a PR is against anyway, no need to recreate PRs).

micwoj92:
This branch has conflicts that must be resolved

PR #739, opened 2019-02-17
Complete objidl

Jarcho:
Splitting up #601 

objidl includes objidlbase inline for whatever reason, so a fair number of things aren't included from the header.


src/um/objidl.rs
```diff
@@ -1,30 +1,113 @@
-// Copyright © 2016-2017 winapi-rs developers
+// Copyright © 2015-2017 winapi-rs developers
```
retep998: There's a conflict, and it's probably due to the removal of the copyright year lines.

src/um/objidl.rs
```diff
@@ -1,29 +1,113 @@
+// Copyright © 2016-2019 winapi-rs developers
```
retep998: What I meant was that I intentionally removed copyright years because they cause a lot of churn and don't actually mean anything legally, so you should remove it too.

Jarcho: Done. I'll keep that in mind for the future.

PR #740, opened 2019-02-17
Complete oaidl

Jarcho:
Splitting up #601


FXTi:
Can't I just use basedef from km?

retep998:
This PR has conflicts that need to be resolved.

Issue #722, opened 2018-12-31
I need IDesktopWallpaper::SetWallpaper


MOZGIII:
Hello, as suggested in the readme, I'm creating this issue.

I need a way to use `IDesktopWallpaper::SetWallpaper` (https://docs.microsoft.com/en-us/windows/desktop/api/shobjidl_core/nf-shobjidl_core-idesktopwallpaper-setwallpaper).

I think I need to use is via COM, but I don't know if it's currently possible, and if it is - how to do it.

Thanks.

retep998:
Once bindings have been added it should be possible. The class `DesktopWallpaper` implements the `IDesktopWallpaper` interface, so you can just use `CoCreateInstance` to get a `IDesktopWallpaper` object and then call the `SetWallpaper` method on it.

MOZGIII:
So, the first step is bindings. Can you give me a code example of how can I create the bindings? Should it be a PR to this repo, or are they supposed to live in my own code base?

https://docs.rs/winapi/0.3.6/winapi/um/combaseapi/fn.CoCreateInstance.html - is this the call you're talking about?

retep998:
Because `IDesktopWallpaper` is defined in the Windows SDK, you would submit a PR to this repo adding it.

It lives in `um/shobjidl_core.h` in the Windows SDK and the equivalent module in this crate already exists and has some examples in it: https://github.com/retep998/winapi-rs/blob/0.3/src/um/shobjidl_core.rs

`CONTRIBUTING.md` has a lot of useful information.

lberrymage:
Since this functionality was merged in #745, I think we can close this issue.

PR #745, opened 2019-03-03
 Add IDesktopWallpaper interface to shobjidl_core. 

AlphaModder:
I needed this interface for a project, and saw that so does @MOZGIII in #722, so I decided to write up the bindings myself. 

Pretty sure I've got everything right, but there is one thing: in the SDK, `IShellItemArray` is defined in `shobjidl_core.h`, but in `winapi` it's under `shobjidl.rs`. Because `IDesktopWallpaper` uses it, currently this PR introduces a dependency on the `shobjidl` feature into the `shobjidl_core` features, which may not be ideal.

Since both leaving `IShellItemArray` where it is and introducing a dependency or moving it to `shobjidl_core.rs` (restoring consistency with the SDK) would result in a breaking change, I suggest that if backwards compatibility is a concern, we move it to `shobjidl_core.rs`, but reexport it in `shobjidl.rs`.


retep998:
>I suggest that if backwards compatibility is a concern, we move it to shobjidl_core.rs, but reexport it in shobjidl.rs.

This is the correct course of action. If a Windows SDK update caused things to be moved, then they should be defined in their new location and re-exported in their old location to prevent breaking changes (until a new major version of `winapi` is published).

AlphaModder:
@retep998 Wow, thanks for the quick reply. I'll set up the reexport immediately.

AlphaModder:
@retep998 Hmm, I've added the reexport, but it puts some new imports in `shobjidl_core` (`propsys` and `propkeydef`), which go unresolved. Do I need to specify a dependency somewhere?

retep998:
You have to update the feature array in `build.rs`.

AlphaModder:
@retep998 Got it! Thanks.

src/um/shobjidl.rs
```diff
@@ -3,60 +3,16 @@
...
-    ) -> HRESULT,
-    fn GetItemAt(
-        dwIndex: DWORD,
-        ppsi: *mut *mut IShellItem,
-    ) -> HRESULT,
-    // TODO: Add IEnumShellItems
-    //fn EnumItems(
-    //    ppenumShellItems: *mut *mut IEnumShellItems,
-    //) -> HRESULT,
-}}
```
retep998, line 59: Don't forget to actually reexport the stuff that was moved!

AlphaModder: Ah, I did it in the `pub use` above. Would a `pub type` be better?

retep998: No, `pub use` is fine. It should go in the spot where the definitions used to exist though, with a `// FIXME` stating to remove them in the next major version.

AlphaModder: Ah, got it.

MOZGIII:
This is great, looking forward to use it!

AlphaModder:
@retep998 Is there anything left to do here, or is this ready to merge?

MOZGIII:
I've implemented my feature using @AlphaModder's branch, and it works for me.
You can take a look here: https://github.com/MOZGIII/wallpaper-rs/tree/master/wallpaper-windows-shobjidl
At this point, whe only thing preventing me from publishing my crate is the dependency on the git branch instead of the proper crates.io-pubished version.
Can we merge this?

retep998 (approved):

PR #758, opened 2019-04-16
add definitions from esent.h

wfraser:
ESE is a low-level database storage engine included in Windows and used by a number of components and apps.

As an example, here's a crate that uses this to build a safer high-level interface, including an example program that reads the Groove Music app's database: https://github.com/wfraser/esedb-rs

This module is not 100% complete with all struct definitions, but I believe I've got all the functions defined and therefore all types referred to by the function signatures are at least there.

Fair warning: I've only tested a handful of functions, and this conversion is largely mechanical, but it's pretty straightforward. I have tested that the alignment of all structs is compatible with and that all struct field offsets match that of the C header.


wfraser:
Seems like Rust 1.6 doesn't like the packing preamble...
```
src\um\esent.rs:59:60: 59:61 error: expected ident, found `4`
src\um\esent.rs:59     #[cfg_attr(target_pointer_width = "32", repr(C, packed(4)))]
                                                                              ^
Could not compile `winapi`.
```

I don't know how to achieve that type of struct layout in 1.6. The `#[repr(packed(N))]` feature was only stabilized in 1.33: https://github.com/rust-lang/rust/commit/87c8cdd8efb60a9a4b57e5b959bc56d36ea511b5

retep998:
Leave out the `packed(N)` attributes for now, and if any type has the wrong size or alignment add a `//FIXME` comment to it so it can be fixed when 0.4 starts getting worked on which will bump the minimum Rust version to 1.33.

wfraser:
I wrote a super ugly program to emit test programs that dump the size, alignment, and offset of each member in each struct in both C++ and in Rust, and verified that actually none of them depend on the `#[repr(packed(N))]` attribute. The only differences were that some of the Rust structs end up over-aligned, but this is safe. So I've removed the attributes on all structs (except the few that are still `#[repr(packed)]`).

PR #760, opened 2019-04-23
Add #[debug] to descriptor structs

### LINKED ISSUE #542: Derive Clone + Debug for descriptor structs
Things like `D3D11_RASTERIZER_DESC` we used to print to debug output in `gfx_device_d3d11`...

tangmi:
Closes #542 

- Add `#[debug]` support for `UNION!`
- Add `#[debug]` to `.*?_DESC\d?` in `(d2d|d3d|dxgi).*`
  - And add `#[debug]` to any types depended on by above.


retep998:
I'm not sure how I feel about `Debug` on unions. I'll think about it and come up with a solution.

Lucretiel:
> I'm not sure how I feel about `Debug` on unions. I'll think about it and come up with a solution.

We have a need for this in https://github.com/crossterm-rs/crossterm/pull/363. Wouldn't a reasonable solution be to just print the underlying `[u16]`?

PR #767, opened 2019-05-06
RFC: Add Windows Hypervisor Platform APIs

benpye:
I'm looking to get some advice on how to get this merged. This adds the [Windows Hypervisor Platform APIs](https://docs.microsoft.com/en-us/virtualization/api/).

There is one type that has been especially troublesome, `WHV_UINT128`. It represents a 128 bit integer. I have had to manually specify the alignment on it and all unions using it to ensure it has the correct 16 byte alignment. Unfortunately `u128` in Rust is not stable over the FFI boundary and in any case, it currently has the wrong alignment, see https://github.com/rust-lang/rust/issues/54341

It seems to me that at least the following still need doing:
- [ ] Structs/unions should have tests to at least ensure they are the expected size/alignment
- [ ] .def file needs writing for non msvc platform - only x86_64 is supported by WHVP.


retep998:
Because `WHV_UINT128` does require `#[repr(align(16))]` I will likely be unable to merge this until work on winapi 0.4 commences.

PR #775, opened 2019-06-11
Trying to add `NtDeviceIoControlFile`

FXTi:
I know this is a deprecated function, but truly useful to communicate with AFD driver: https://docs.microsoft.com/en-us/windows/desktop/api/winternl/nf-winternl-ntdeviceiocontrolfile

Review is needed since I'm not so confident about the correctness. I'm fully prepared to multiple commits. : )


FXTi:
Can't I just use basedef from km?

PR #782, opened 2019-07-04
Changed handleapi feature link library from kernel32 to WindowsApp to…

### LINKED ISSUE #781: Link error with CompareObjectHandles
Using winapi::um::handleapi::CompareObjectHandles causes a link error to appear:
```
 = note: lld-link: error: undefined symbol: CompareObjectHandles
          >>> referenced by C:\git-mozilla\servo_ipc-channel\ipc-channel\src\platform\windows\mod.rs:349
          >>>               C:\git-mozilla\servo_ipc-channel\ipc-channel\target\debug\deps\ipc_channel-d21c39ad54b8c1b9.grgdvyw18xi6bb.rcgu.o:(_ZN82_$LT$ipc_channel..platform..windows..WinHandle$u20$as$u20$core..cmp..PartialEq$GT$2eq17h4f6b02f15a66ad57E)
```
I suspect this is because it's provided from kernelbase.lib instead of kernel32.lib, per https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-compareobjecthandles.

angelortiz1007:
While doing builds of ipc-channel and using windows 10 api CompareObjectHandles() I received an unresolved message for CompareObjectHandles().

Reading Microsoft link: https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis I changed build.rs, handleapi link time library from kernel32 to WindowsApp.

After testing locally on my ipc-channel build and using my patched winapi v0.3.0 I removed the link time unresolved error and all ipc-channel tests passed.

… address/fix https://github.com/retep998/winapi-rs/issues/781.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -195,7 +195,7 @@ const DATA: &'static [(&'static str, &'static [&'static str], &'static [&'static
     ("fileapi", &["minwinbase", "minwindef", "winnt"], &["kernel32"]),
     ("functiondiscoverykeys_devpkey", &["wtypes"], &[]),
     ("gl-gl", &[], &["opengl32"]),
-    ("handleapi", &["minwindef", "winnt"], &["kernel32"]),
+    ("handleapi", &["minwindef", "winnt"], &["WindowsApp"]),
     ("heapapi", &["basetsd", "minwinbase", "minwindef", "winnt"], &["kernel32"]),
     ("highlevelmonitorconfigurationapi", &["minwindef", "physicalmonitorenumerationapi", "winnt"], &["dxva2"]),
     ("http", &["guiddef", "minwinbase", "minwindef", "sspi", "winnt", "ws2def"], &["winhttp"]),


retep998:
If `WindowsApp.lib` were linked to, depending on order in the linker many functions will end up resolving to various `api-ms-win-core` dlls rather than older dlls. For example `GetLocalTime` could end up coming from `api-ms-win-core-sysinfo-l1-1-0.dll` instead of `kernel32.dll`. This will absolutely break support for older Windows that do not have the `api-ms-win-core` dlls such as Windows 7.

If we had the [raw-dylib](https://github.com/rust-lang/rust/issues/58713) feature then this would be a much easier problem to solve but unfortunately we won't for a long time.

I'm not entirely sure how to solve this issue, and for now my recommendation is to just load the function via `GetProcAddress` or a crate that wraps that.

jdm:
Good to know that that's a possible workaround. Thanks!

angelortiz1007:
The LoadLibrary() of api-ms-win-core-sysinfo-l1-1-0.dll followed by GetProcAddress() should work.   

Obviously, if ipc-channel - which uses CompareObjectHandles() - is mondified to use LoadLibrary()/GetProcAddress() 1 test will still fail if ipc-channel is used on non-windows 10 systems.

retep998:
Because this leads to other undesired changes, this PR cannot be merged.

PR #792, opened 2019-07-30
Add IsWow64Process2

lzybkr:
IsWow64Process use is discouraged, favoring IsWow64Process2 which was missing in winapi-rs.


retep998 (changes requested):
`IsWow64Process2` should come after `GetSystemWow64DirectoryW`.

retep998 (approved):

PR #794, opened 2019-08-06
Add tip on how to use FARPROC pointers

larsch:


README.md
```diff
@@ -45,6 +45,10 @@ Yes, absolutely! By default the `std` feature of `winapi` is disabled, allowing
 
 Because `winapi` does not depend on `std` by default, it has to define `c_void` itself instead of using `std::os::raw::c_void`. However, if you enable the `std` feature of `winapi` then it will re-export `c_void` from `std` and cause `winapi`'s `HANDLE` to be the same type as `std`'s `HANDLE`.
 
+### How do I turn `FARPROC` pointers (e.g. from `GetProcAddress`) into callable functions?
+
+Using `let func: Option<extern "stdcall" fn()> = unsafe { std::memory::transmute_copy(&ptr) };`.  According to [The Rustnomicon](https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization), `Option<fn()>` is the correct way to represent a nullable function pointer, corresponding to `void (*)()` in C.
```
Lokathor, line 50: This should be `transmute`, not `transmute_copy`.

`transmute_copy` is even more "last resort" than a normal `transmute`, and should not be required to turn a `*mut c_void` into a `Option<fn>`. They'll be the same size, and legal for a standard `transmute`.

Lokathor (changes requested):

README.md
retep998:
```suggestion
Using `let func: Option<extern "stdcall" fn()> = unsafe { std::mem::transmute(ptr) };`.  According to [The Rustnomicon](https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization), `Option<fn()>` is the correct way to represent a nullable function pointer, corresponding to `void (*)()` in C.
```

retep998 (changes requested):

PR #798, opened 2019-08-14
Format code using 'cargo fmt'

Atul9:


MSxDOS:
I'd suggest reading [CONTRIBUTING.md](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md) first.

retep998:
I do want to use rustfmt but there are several things I'd need from it first before I can use it for `winapi`. When rustfmt is ready to be used for `winapi` I'll do the changes myself.

PR #801, opened 2019-08-26
Fix deprecated warnings in test

hankbao:
This PR fixes the deprecated warnings about the using of `trim_left` and `trim_right` in the test code.


hankbao:
If this crate needs to support gnu-1.6.0, feel free to close this PR then.

retep998:
These deprecation warnings will be fixed in winapi 0.4 when the minimum Rust version is bumped.

PR #772, opened 2019-06-02
Iphlpapi addition

BlackHoleFox:
All cargo tests pass and it builds on stable with no issues.

Follow up to #744 that I closed because I messed up the Git history pretty badly.

Issue #676, opened 2018-09-02
Question: Handling of #define SIZEOF_X Functions


mattlknight:
I'm still very new to C and FFI, as well as this project. Are we supposed to be adding in the rust equivalent for the #DEFINE functions that provide a size_of return value?

For example, this is a define in ifmib.h
```c
#define SIZEOF_IFTABLE(X) (FIELD_OFFSET(MIB_IFTABLE,table[0]) + \
                           ((X) * sizeof(MIB_IFROW)) + ALIGN_SIZE)
```

The full block is this:
```c
typedef struct _MIB_IFTABLE {
    DWORD dwNumEntries;
    MIB_IFROW table[ANY_SIZE];
} MIB_IFTABLE, *PMIB_IFTABLE;

#define SIZEOF_IFTABLE(X) (FIELD_OFFSET(MIB_IFTABLE,table[0]) + \
                           ((X) * sizeof(MIB_IFROW)) + ALIGN_SIZE)
```

I don't really understand if I should be defining some sort of function to return the value that SIZEOF_IFTABLE(Z) would, or if I should be calculating it in advance and storing it as a pub const number.

Can someone provide some input on this? If I am supposed to be defining this, can you provide/point me to an example?

Thanks,
Matt

retep998:
Just leave a FIXME with that commented out. While `size_of` is already a nuisance due to not being a const function yet, the bigger hurdle is `FIELD_OFFSET` which requires `offset_of` which doesn't exactly exist yet.

mattlknight:
Thanks will do!

Inc0n:
Hi, I am fairly new to rust, and trying to find std::mem::size_of in place of c's sizeof on PROCESSENTRY32,
but it's complaining mismatched type
```
error[E0308]: mismatched types  
  |  
  |   procEntry.dwSize = std::mem::size_of::<PROCESSENTRY32>(); 
  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u32,  found usize  
```

How do i fix this? 
Thanks in advance

retep998:
Rust does not do implicit numeric conversions unlike C so you have to cast the type.
`procEntry.dwSize = std::mem::size_of::<PROCESSENTRY32>() as DWORD;`

Inc0n:
ah, thanks, that fix it for me 😅

PR #802, opened 2019-08-26
iphlpapi support

hankbao:
It looks like the #772 has been stalled for a while. So I'd like to give it a try.


hankbao:
Added SIZEOF_* functions

hankbao:
@retep998 Hi, I've fixed a bunch of issues and added necessary tests. Now all tests passed. Please take some time to review this PR. Thanks.

src/shared/ipmib.rs
```diff
@@ -0,0 +1,396 @@
...
+use shared::ifdef::IF_INDEX;
+use shared::ifmib::MAXLEN_PHYSADDR;
+use shared::minwindef::DWORD;
+use shared::nldef::NL_ROUTE_PROTOCOL;
+use shared::ntdef::{PVOID, UCHAR, ULONG, USHORT};
+use um::rtinfo::ALIGN_SIZE;
+const ANY_SIZE: usize = 1;
+macro_rules! offset_of_table {
+    ($container:ty) => (unsafe {
+        &(*(ptr::null() as *const $container)).table[0] as *const _ as usize
```
retep998: This is UB because it dereferences a null pointer.

src/shared/ntddndis.rs
```diff
@@ -0,0 +1,59 @@
...
+    NdisMediumArcnetRaw = 6,
+    NdisMediumArcnet878_2 = 7,
+    NdisMediumAtm = 8,
+    NdisMediumWirelessWan = 9,
+    NdisMediumIrda = 10,
+    NdisMediumBpc = 11,
+    NdisMediumCoWan = 12,
+    NdisMedium1394 = 13,
+    NdisMediumInfiniBand = 14,
+// #if ((NTDDI_VERSION >= NTDDI_VISTA) || NDIS_SUPPORT_NDIS6)
```
retep998: Remove these comments

src/shared/tcpmib.rs
```diff
@@ -0,0 +1,283 @@
...
+use shared::basetsd::DWORD64;
+use shared::in6addr::IN6_ADDR;
+use shared::minwindef::DWORD;
+use shared::ntdef::{LARGE_INTEGER, UCHAR, ULONGLONG};
+use um::rtinfo::ALIGN_SIZE;
+pub const TCPIP_OWNING_MODULE_SIZE: usize = 16;
+const ANY_SIZE: usize = 1;
+macro_rules! offset_of_table {
+    ($container:ty) => (unsafe {
+        &(*(ptr::null() as *const $container)).table[0] as *const _ as usize
```
retep998: See above regarding UB.

src/shared/tcpmib.rs
```diff
@@ -0,0 +1,283 @@
...
+    dwInErrs: DWORD,
+    dwOutRsts: DWORD,
+    dwNumConns: DWORD,
+}}
+pub type PMIB_TCPSTATS_W2K = *mut MIB_TCPSTATS_W2K;
+// Windows Vista
+// pub type MIB_TCPSTATS = MIB_TCPSTATS_LH;
+pub type MIB_TCPSTATS = MIB_TCPSTATS_W2K;
+pub type PMIB_TCPSTATS = *mut MIB_TCPSTATS;
+// if NTDDI_VERSION >= NTDDI_WIN10_RS3
```
retep998: Remove these comments.

src/shared/udpmib.rs
```diff
@@ -0,0 +1,160 @@
...
+use shared::basetsd::DWORD64;
+use shared::in6addr::IN6_ADDR;
+use shared::minwindef::DWORD;
+use shared::ntdef::{INT, LARGE_INTEGER, UCHAR, ULONGLONG};
+use um::rtinfo::ALIGN_SIZE;
+const ANY_SIZE: usize = 1;
+pub const TCPIP_OWNING_MODULE_SIZE: usize = 16;
+macro_rules! offset_of_table {
+    ($container:ty) => (unsafe {
+        &(*(ptr::null() as *const $container)).table[0] as *const _ as usize
```
retep998: See above regarding UB.

src/um/ipexport.rs
```diff
@@ -0,0 +1,174 @@
...
+    OptionsData: PUCHAR,
+}}
+pub type PIP_OPTION_INFORMATION = *mut IP_OPTION_INFORMATION;
+#[cfg(target_arch = "x86_64")]
+STRUCT!{struct IP_OPTION_INFORMATION32 {
+    Ttl: UCHAR,
+    Tos: UCHAR,
+    Flags: UCHAR,
+    OptionsSize: UCHAR,
+    OptionsData: u32, // UCHAR * POINTER_32
```
retep998, line 32: `__ptr32` is so weird...

src/um/iphlpapi.rs
```diff
@@ -0,0 +1,524 @@
...
+    Flags: ULONG,
+    SystemTimestamp1: ULONG64,
+    HardwareClockTimestamp: ULONG64,
+    SystemTimestamp2: ULONG64,
+}}
+pub type PINTERFACE_HARDWARE_CROSSTIMESTAMP = *mut INTERFACE_HARDWARE_CROSSTIMESTAMP;
+STRUCT!{struct HIFTIMESTAMPCHANGE__ {
+    unused: i32,
+}}
+pub type HIFTIMESTAMPCHANGE = *mut HIFTIMESTAMPCHANGE__;
```
retep998: This should be defined with `DECLARE_HANDLE!`

retep998 (changes requested):

src/shared/ipmib.rs
hankbao: I'm not sure if I understand what UB means correctly. But this is actually doing pointer arithmetic and give out the offset of `table` and I believe it [runs well](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=fce1a2aad45bd8b8003c467342bd50b5)

Could you please elaborate on the matter?

src/um/ipexport.rs
hankbao: Yea, I can't figure out how to do it in rust. So I just leave a `u32` here.

src/shared/ipmib.rs
retep998: UB is Undefined Behavior. Dereferencing a null pointer is undefined behavior. Just because it currently happens to do the right thing does not mean it is guaranteed to behave correctly and it could cause horrible things to happen at any time depending on LLVM optimizations.

hankbao: If you're referencing the [rust reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), then yes, dereferences a null pointer is listed on it.

Do you think it's acceptable if I do something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2ff6cbe8b43643ec91b45dfae2e8799c)

MSxDOS: There's currently no way in Rust to take an offset of a field without UB, so people (including myself) who implement their own ways of doing it are doing so at their own risk.
See https://github.com/Gilnaa/memoffset/issues/24 and https://github.com/rust-lang/rfcs/pull/2582

hankbao: > There's currently no way in Rust to take an offset of a field without UB, so people (including myself) who implement their own ways of doing it are doing so at their own risk.
> See [Gilnaa/memoffset#24](https://github.com/Gilnaa/memoffset/issues/24) and [rust-lang/rfcs#2582](https://github.com/rust-lang/rfcs/pull/2582)

Thanks for pointing out the `offset_of` discussion.

But is it UB even with doing pointer arithmetic of a `mem::zeroed() T` in this particular scenario like: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2ff6cbe8b43643ec91b45dfae2e8799c ?

hankbao: If there is no defined way to do `offset_of`, maybe I could replace all these `SIZEOF_*` 
 with some `FIXME` comments as @retep998 mentioned in #676?

hankbao: I've replaced all `SIZEOF_*` in this PR with `FIXME` comments.

src/um/ipexport.rs
hankbao: Since there is nothing we can do about `__ptr32`, mark this as resolved.

hankbao:
@retep998 Please continue your review. Thanks.

src/shared/ipmib.rs
MSxDOS: > But is it UB even with doing pointer arithmetic of a `mem::zeroed() T` in this particular scenario like:

Should be fine for winapi structs as `zeroed` is a valid state there.

retep998: I'd still rather they just remain commented out for now and later I'll figure out what I want to do with them.

hankbao: > > But is it UB even with doing pointer arithmetic of a `mem::zeroed() T` in this particular scenario like:
> 
> Should be fine for winapi structs as `zeroed` is a valid state there.

I guess so. I've commented out these `SIZEOF_*`s anyway.

> I'd still rather they just remain commented out for now and later I'll figure out what I want to do with them.

Sure.

TyPR124:
Is there anything here that still needs changed/fixed here, or just waiting on review?

Also I am not seeing the [ICMP API](https://docs.microsoft.com/en-us/windows/win32/api/icmpapi/nf-icmpapi-icmpsendecho) in here. Unless someone else plans to or has already done this, I can work on that once this is merged.

hankbao:
Still waiting for this PR to get reviewed.

src/um/ipexport.rs
```diff
@@ -0,0 +1,174 @@
...
+// except according to those terms.
+// #include <winapifamily.h>
+// #include <in6addr.h>
+// #include <inaddr.h>
+use shared::basetsd::ULONG64;
+use shared::in6addr::in6_addr;
+use shared::ntdef::{INT, PUCHAR, PVOID, UCHAR, ULONG, USHORT, WCHAR};
+pub const MAX_ADAPTER_NAME: usize = 128;
+pub const MAX_OPT_SIZE: usize = 40;
+pub type IPAddr = ULONG;
```
TyPR124, line 14: I wanted to point out that I think this definition of IPAddr may cause issues. This allows one to write 

```rust
let rust_ip = Ipv4Addr::new(1,2,3,4);
let win_ip: IPAddr = rust_ip.into();
```

However this actually swaps the byteorder. A packet sent to the `win_ip` value will go to 4.3.2.1 instead of 1.2.3.4.

retep998: Regardless, it is the correct definition, and you just have to be really careful about byte order when working with windows API.

TyPR124: I don't disagree that it is "correct", but perhaps it is worth adding a small doc comment or something?

I was (until just now) actually under the impression that IPAddr should have been an alias for IN_ADDR, considering everywhere I see IPAddr in MS documentation, it links to IN_ADDR instead. Looking at the header file, I see I was incorrect, though docs do say that the types are compatible.

I'm just concerned that this is a very easy mistake to make outside of an unsafe block and there is no documentation that states this is a problem.

hankbao: > ```rust
> let rust_ip = Ipv4Addr::new(1,2,3,4);
> let win_ip: IPAddr = rust_ip.into();
> ```
> 
> However this actually swaps the byteorder. A packet sent to the `win_ip` value will go to 4.3.2.1 instead of 1.2.3.4.

> I was (until just now) actually under the impression that IPAddr should have been an alias for IN_ADDR, considering everywhere I see IPAddr in MS documentation, it links to IN_ADDR instead. Looking at the header file, I see I was incorrect, though docs do say that the types are compatible.

This issue has nothing to do with MSDN docs. You may check the rust docs and find out things like this:

```rust
impl From<Ipv4Addr> for u32
fn from(ip: Ipv4Addr) -> u32
Converts an Ipv4Addr into a host byte order u32.
```

`rust_ip.into()` produces a host byte order u32 and you're likely working on a little-endian machine.

TyPR124: I'll accept that nothing will be done to address this; so be it. I still feel very uncomfortable about the fact that

```rust
let ip = IPAddr::from(Ipv4Addr::new(...));
```

looks like completely correct code, but it is not. It's not obvious that u32 is even involved. Nor does any bit of documentation make this obvious. The *only* way you know this is invalid is to actually look at the definition of IPAddr and see that it is actually u32.

> you're likely working on a little-endian machine.

As far as I know, Windows only runs on little endian so this is a potential issue for every user, which is why I even thought bring this up. I still think a small doc comment would go a long way, but again I accept that this won't happen.

ErichDonGubler:
@hankbao @retep998: Anything stopping this from getting merged? I'd really like to see https://github.com/heim-rs/heim/issues/29 resolved and am not sure how to help!

hankbao:
Need to get reviewed by @retep998

GuillaumeGomez:
@retep998 Need a hand?

retep998:
@GuillaumeGomez I've been far too busy to give `winapi` the attention it deserves, so I'd be happy to have some help.

GuillaumeGomez:
Ok, let's talk about it on IRC whenever we're both present at the same time!

vkrasnov:
Hi all, this would be very useful to me. I can help reviewing if that helps.

yuguorui:
Please...

src/shared/ifdef.rs
```diff
@@ -3,20 +3,198 @@
...
+    NET_IF_ADMIN_STATUS_UP = 1,
+    NET_IF_ADMIN_STATUS_DOWN = 2,
+    NET_IF_ADMIN_STATUS_TESTING = 3,
+}}
+pub type PNET_IF_ADMIN_STATUS = *mut NET_IF_ADMIN_STATUS;
+pub type NET_IF_COMPARTMENT_SCOPE = UINT32;
+pub type PNET_IF_COMPARTMENT_SCOPE = *mut NET_IF_COMPARTMENT_SCOPE;
+pub const NET_IF_COMPARTMENT_SCOPE_UNSPECIFIED: NET_IF_COMPARTMENT_SCOPE = 0;
+pub const NET_IF_COMPARTMENT_SCOPE_ALL: NET_IF_COMPARTMENT_SCOPE = -1i32 as
+    NET_IF_COMPARTMENT_SCOPE;
```
retep998:
```suggestion
pub const NET_IF_COMPARTMENT_SCOPE_ALL: NET_IF_COMPARTMENT_SCOPE = -1i32 as u32;
```
As per https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#constants

src/shared/ifdef.rs
```diff
@@ -3,20 +3,198 @@
...
+STRUCT!{struct NET_PHYSICAL_LOCATION_LH {
+    BusNumber: ULONG,
+    SlotNumber: ULONG,
+    FunctionNumber: ULONG,
+}}
+pub type PNET_PHYSICAL_LOCATION_LH = *mut NET_PHYSICAL_LOCATION_LH;
+pub const IF_MAX_STRING_SIZE: usize = 256;
+pub const IF_MAX_PHYS_ADDRESS_LENGTH: usize = 32;
+STRUCT!{struct IF_COUNTED_STRING_LH {
+    Length: USHORT, // in -Bytes-
```
retep998:
```suggestion
    Length: USHORT,
```

src/shared/ifdef.rs
```diff
@@ -3,20 +3,198 @@
...
+pub type PIF_ADMINISTRATIVE_STATE = *mut IF_ADMINISTRATIVE_STATE;
+ENUM!{enum IF_OPER_STATUS {
+    IfOperStatusUp = 1,
+    IfOperStatusDown = 2,
+    IfOperStatusTesting = 3,
+    IfOperStatusUnknown = 4,
+    IfOperStatusDormant = 5,
+    IfOperStatusNotPresent = 6,
+    IfOperStatusLowerLayerDown = 7,
+}}
```
retep998:
```suggestion
ENUM!{enum IF_OPER_STATUS {
    IfOperStatusUp = 1,
    IfOperStatusDown,
    IfOperStatusTesting,
    IfOperStatusUnknown,
    IfOperStatusDormant,
    IfOperStatusNotPresent,
    IfOperStatusLowerLayerDown,
}}
```
Enum definitions should match the headers, and in the header only the value for the first variant is specified.
Check the other enums to make sure they match as well.

src/shared/netioapi.rs
```diff
@@ -3,14 +3,552 @@
...
+    InterfaceIndex: NET_IFINDEX,
+    InterfaceGuid: GUID,
+    Alias: [WCHAR; IF_MAX_STRING_SIZE + 1],
+    Description: [WCHAR; IF_MAX_STRING_SIZE + 1],
+    PhysicalAddressLength: ULONG,
+    PhysicalAddress: [UCHAR; IF_MAX_PHYS_ADDRESS_LENGTH],
+    PermanentPhysicalAddress: [UCHAR; IF_MAX_PHYS_ADDRESS_LENGTH],
+    Mtu: ULONG,
+    Type: ULONG, // Interface Type.
+    TunnelType: TUNNEL_TYPE, // Tunnel Type, if Type = IF_TUNNEL.
```
retep998:
```suggestion
    Type: ULONG,
    TunnelType: TUNNEL_TYPE,
```

src/shared/netioapi.rs
```diff
@@ -3,14 +3,552 @@
...
+    MibParameterNotification = 0,
+    MibAddInstance = 1,
+    MibDeleteInstance = 2,
+    MibInitialNotification = 3,
+}}
+pub type PMIB_NOTIFICATION_TYPE = *mut MIB_NOTIFICATION_TYPE;
+STRUCT!{struct MIB_IF_ROW2_InterfaceAndOperStatusFlags {
+    bitfield: BYTE,
+}}
+BITFIELD!{MIB_IF_ROW2_InterfaceAndOperStatusFlags bitfield: BYTE [
```
retep998:
```suggestion
BITFIELD!{MIB_IF_ROW2_InterfaceAndOperStatusFlags bitfield: BOOLEAN [
```
Make sure to match the types the headers use.

src/shared/nldef.rs
```diff
@@ -0,0 +1,204 @@
...
+pub type PNL_NETWORK_CATEGORY = *mut NL_NETWORK_CATEGORY;
+ENUM!{enum NL_INTERFACE_NETWORK_CATEGORY_STATE {
+    NlincCategoryUnknown = 0,
+    NlincPublic = 1,
+    NlincPrivate = 2,
+    NlincDomainAuthenticated = 3,
+    NlincCategoryStateMax = 4,
+}}
+pub type PNL_INTERFACE_NETWORK_CATEGORY_STATE = *mut NL_INTERFACE_NETWORK_CATEGORY_STATE;
+pub const NET_IF_CURRENT_SESSION : ULONG = -1i32 as u32;
```
retep998:
```suggestion
pub const NET_IF_CURRENT_SESSION: ULONG = -1i32 as u32;
```

retep998 (changes requested):
Sorry for taking so long to get to this.

src/shared/netioapi.rs
```diff
@@ -3,14 +3,552 @@
...
+    InterfaceLuid: NET_LUID,
+    InterfaceIndex: NET_IFINDEX,
+    InterfaceGuid: GUID,
+    Alias: [WCHAR; IF_MAX_STRING_SIZE + 1],
+    Description: [WCHAR; IF_MAX_STRING_SIZE + 1],
+    PhysicalAddressLength: ULONG,
+    PhysicalAddress: [UCHAR; IF_MAX_PHYS_ADDRESS_LENGTH],
+    PermanentPhysicalAddress: [UCHAR; IF_MAX_PHYS_ADDRESS_LENGTH],
+    Mtu: ULONG,
+    Type: ULONG, // Interface Type.
```
retep998:
```suggestion
    Type: IFTYPE,
```

src/shared/nldef.rs
```diff
@@ -0,0 +1,204 @@
...
+    IpSuffixOriginLinkLayerAddress = 4,
+    IpSuffixOriginRandom = 5,
+    IpSuffixOriginUnchanged = 1 << 4,
+}}
+pub const NlsoOther: NL_SUFFIX_ORIGIN = IpSuffixOriginOther;
+pub const NlsoManual: NL_SUFFIX_ORIGIN = IpSuffixOriginManual;
+pub const NlsoWellKnown: NL_SUFFIX_ORIGIN = IpSuffixOriginWellKnown;
+pub const NlsoDhcp: NL_SUFFIX_ORIGIN = IpSuffixOriginDhcp;
+pub const NlsoLinkLayerAddress: NL_SUFFIX_ORIGIN = IpSuffixOriginLinkLayerAddress;
+pub const NlsoRandom: NL_SUFFIX_ORIGIN = IpSuffixOriginRandom;
```
retep998:
```suggestion
ENUM!{enum NL_SUFFIX_ORIGIN {
    NlsoOther = 0,
    NlsoManual,
    NlsoWellKnown,
    NlsoDhcp,
    NlsoLinkLayerAddress,
    NlsoRandom,
    IpSuffixOriginOther = 0,
    IpSuffixOriginManual,
    IpSuffixOriginWellKnown,
    IpSuffixOriginDhcp,
    IpSuffixOriginLinkLayerAddress,
    IpSuffixOriginRandom,
    IpSuffixOriginUnchanged = 1 << 4,
}}
```
Make sure to match what the headers do.

src/um/iptypes.rs
```diff
@@ -0,0 +1,379 @@
...
+    Dhcpv4Enabled set_Dhcpv4Enabled[2..3],
+    ReceiveOnly set_ReceiveOnly[3..4],
+    NoMulticast set_NoMulticast[4..5],
+    Ipv6OtherStatefulConfig set_Ipv6OtherStatefulConfig[5..6],
+    NetbiosOverTcpipEnabled set_NetbiosOverTcpipEnabled[6..7],
+    Ipv4Enabled set_Ipv4Enabled[7..8],
+    Ipv6Enabled set_Ipv6Enabled[8..9],
+    Ipv6ManagedAddressConfigurationSupported set_Ipv6ManagedAddressConfigurationSupported[9..10],
+    Reserved set_Reserved[10..32],
+]}
```
retep998: As this is an *anonymous* union of a single field and a corresponding bitfield, this should follow https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#union-with-a-primitive-field-and-an-anonymous-bitfield-struct-of-the-same-type and have the bitfield applied directly to a `Flags` field in `IP_ADAPTER_ADDRESSES_LH` without the intermediary `IP_ADAPTER_ADDRESSES_LH_s`.

hankbao:
@retep998 all issues resolved. Please continue your review. Hopefully, this PR will get merged shortly.

Issue #743, opened 2019-03-01
http module needs httpapi.lib, not winhttp.lib


tom-bowles:
I think this line in build.rs:

    ("http", &["guiddef", "minwinbase", "minwindef", "sspi", "winnt", "ws2def"], &["winhttp"]),

should be:

    ("http", &["guiddef", "minwinbase", "minwindef", "sspi", "winnt", "ws2def"], &["httpapi"]),

jackyzy823:
When I try to use httpapi, I found this question too.

And another question is :
https://github.com/retep998/winapi-rs/blob/f823e7e876d569aee316962ff7262c89f26c72df/src/um/http.rs#L853-L856

According to https://docs.microsoft.com/en-us/windows/win32/api/http/nf-http-httpcreatehttphandle
```c++
HTTPAPI_LINKAGE ULONG HttpCreateHttpHandle(
  OUT PHANDLE RequestQueueHandle,
  ULONG       Reserved
);
```
pReqQueueHandle shoud be `PHANDLE` not `HANDLE`

It is highly appreciated if you could fix these.Thanks

PR #806, opened 2019-08-30
Fix and complete httpapi

jackyzy823:
Fix issues from #743
1. fix library dependency
2. fix wrong parameter type
3. add missing functions and macros

### DIFF
--- a/src/um/http.rs
+++ b/src/um/http.rs
@@ -840,6 +840,15 @@ STRUCT!{struct HTTP_SERVICE_CONFIG_CACHE_SET {
     ParamDesc: HTTP_SERVICE_CONFIG_CACHE_PARAM,
 }}
 pub type PHTTP_SERVICE_CONFIG_CACHE_SET = *mut HTTP_SERVICE_CONFIG_CACHE_SET;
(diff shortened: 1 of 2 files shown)


src/um/http.rs
```diff
@@ -851,7 +860,7 @@ extern "system" {
         pReserved: PVOID,
     ) -> ULONG;
     pub fn HttpCreateHttpHandle(
-        pReqQueueHandle: HANDLE,
+        pReqQueueHandle: PHANDLE,
```
Connicpu, line 863: Even though this is theoretically a breaking change, it would actually be impossible to be correctly using this API as it was without crashing. So looks good to me 👍

Connicpu (approved):
Looks good to me, as long as everything checks out with that one thing I left a comment on

src/um/http.rs
retep998: Breaking changes are allowed when they are correcting bindings.

Connicpu: Cool, then my approval of it stands, looks good to merge~

retep998 (approved):

PR #810, opened 2019-09-17
Add MessageBoxTimeoutA and MessageBoxTimeoutW

alexchandel:
Add the missing, undocumented functions MessageBoxTimeoutA and MessageBoxTimeoutW, present since XP.

### DIFF
--- a/src/um/winuser.rs
+++ b/src/um/winuser.rs
@@ -4669,6 +4669,22 @@ extern "system" {
         uType: UINT,
         wLanguageId: WORD,
     ) -> c_int;
+    pub fn MessageBoxTimeoutA(
+        hWnd: HWND,
+        lpText: LPCSTR,
+        lpCaption: LPCSTR,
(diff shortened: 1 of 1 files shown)


retep998:
I cannot find these in the Windows SDK headers. Is there any official source for these definitions?

alexchandel:
They're undocumented, but they've been present and functional since XP, [and used since then](https://www.codeproject.com/Articles/7914/MessageBoxTimeout-API).  In fact the standard MessageBox functions actually call MessageBoxTimeOutA / MessageBoxTimeoutW under the hood and pass a timeout of 0xFFFFFFFF.

retep998:
Regardless, the policy of `winapi` is to only include bindings that have official headers from Microsoft.

alexchandel:
These are useful functions though, used in the wild, which MS uses in its own implementations.  How about marking them with a warning?

PR #817, opened 2019-10-10
Add meager crate-level documentation

alexchandel:
Add crate-level documentation and mention the current feature gates (which are completely undocumented at the moment).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,16 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
 // All files in the project carrying such notice may not be copied, modified, or distributed
 // except according to those terms.
+//! Raw FFI bindings to all of Windows API.
+//! # Features
+//! Every submodule of this crate's modules lies behind a compiler feature of the same name.
+//! The only exceptions are [`winapi::shared::guiddef`](shared/guiddef/) which requires no feature,
+//! and [`winapi::um::gl`](um/gl/) which instead requires the feature `gl_gl` for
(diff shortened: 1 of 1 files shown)


Lokathor (approved):

alexchandel:
Good to merge?

src/lib.rs
```diff
@@ -3,6 +3,16 @@
 // <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
 // All files in the project carrying such notice may not be copied, modified, or distributed
 // except according to those terms.
+//! Raw FFI bindings to all of Windows API.
+//! # Features
+//! Every submodule of this crate's modules lies behind a compiler feature of the same name.
+//! The only exceptions are [`winapi::shared::guiddef`](shared/guiddef/) which requires no feature,
```
retep998, line 9: There is still a feature for this module, but the module doesn't use it due to the top level `Class`/`Interface` traits needing `GUID`. I'd prefer to leave out the bit about `guiddef` since everything works fine if users treat it the same as other modules.

src/lib.rs
```diff
@@ -3,6 +3,16 @@
...
+//! and [`winapi::um::gl`](um/gl/) which instead requires the feature `gl_gl` for
+//! [`winapi::um::gl::gl`](um/gl/gl/).  Enabling a feature enables all its dependencies.
+//! Additionally, the `everything` feature enables all features, and the `std` feature redefines
```
retep998, line 12: It doesn't enable *all* features, only the module features.

Issue #771, opened 2019-05-16
use core c_void type


Lokathor:
`winapi::ctypes::c_void` should simply be an alias for `core::ffi::c_void`, not a distinct type

retep998:
This is one of the changes that will be happening for version 0.4

Issue #323, opened 2016-08-22
Figure out how to support non-desktop families


retep998:
Currently winapi only supports the desktop family. However for anyone who wants to make a Windows app for a family other than desktop, they will need to have access to a different set of APIs, including linking to different import libraries.

I cannot use cargo features to select which family to use as cargo features are not mutually exclusive https://github.com/rust-lang/cargo/issues/2980

Furthermore, unless the user limits themselves to no_std, they will need the collaboration of libstd as well, as it depends on system functionality which may not exist in the desired family, or may link in an import library which is for the wrong family.

docbrown:
Just brainstorming here, because I'm not very familiar with UWP... What about adding a new target to Rust for UWP apps (say, `x86_64-pc-uwp-msvc`, for example) and then treating `target_os = "windows"` as `WINAPI_PARTITION_DESKTOP` and `target_os = "uwp"` as `WINAPI_PARTITION_APP`?

Susurrus:
Note that rust-lang/rust#60260 is about to be merged which adds the `i686-uwp-windows-gnu` and `x86_64-uwp-windows-gnu` targets. These have the `target_vendor` set to `uwp`, which can be easily filtered on by `cfg` attributes.

PR #831, opened 2019-11-04
add rustc-dep-of-std feature

euclio:
This commit paves the way for libstd to depend on winapi and crates that
depend on winapi.

Tested with a local build of libstd on both `-msvc` and `-gnu` targets.

While it's possible to move small parts of libstd to winapi with this commit, it will be tricky to do a large-scale refactor until #771 and #323 are addressed.


retep998:
This PR fails to build on Rust 1.6 due to cargo changes so this may have to be delayed until winapi 0.4.

euclio:
@retep998 Do you have an idea of what features you'd like to include in a 0.4 release? It would be helpful to have a tracking issue. I'm happy to help push that forward.

retep998:
https://github.com/retep998/winapi-rs/issues/885

PR #840, opened 2019-11-18
icmpapi support

TyPR124:
Builds on #802 to add ICMP API support.


src/um/icmpapi.rs
```diff
@@ -0,0 +1,77 @@
...
+        RequestSize: WORD,
+        RequestOptions: PIP_OPTION_INFORMATION,
+        ReplyBuffer: LPVOID,
+        ReplySize: DWORD,
+        Timeout: DWORD
+    ) -> DWORD;
+    pub fn IcmpSendEcho2Ex(
+        IcmpHandle: HANDLE,
+        Event: HANDLE,
+        ApcRoutine: FARPROC, // or PIO_APC_ROUTINE
```
TyPR124, line 44: Meant to point this out originally... doing it now while I'm thinking about it.
The type of ApcRoutine should be PIO_APC_ROUTINE instead of FARPROC (for any OS after XP), however this type doesn't exist, and implementing it seems to be problematic. The (C) definition is ```typedef
VOID WINAPI
(*PIO_APC_ROUTINE) (
    IN PVOID ApcContext,
    IN PIO_STATUS_BLOCK IoStatusBlock,
    IN ULONG Reserved
    );``` but as far as I can see (P)IO_STATUS_BLOCK is not a type that is defined in this API, but instead comes from a kernel/driver API.

Because of this I left the type as FARPROC, although I could see changing it to PVOID or defining some dummy PIO_APC_ROUTINE.

Same is true in `Icmp6SendEcho2`

PR #841, opened 2019-11-28
IPersistFile, IShellLinkA and IShellLinkW interfaces

TechPriest:
IPersistFile, IShellLinkA and IShellLinkW interfaces plus some flags for IShellLink


TechPriest:
Fixing failed checks

PR #650, opened 2018-07-28
Add wtsapi32.

ekse:
Based on 10.0.17134.0\um\WtsApi32.h.

PR #854, opened 2019-12-24
wtsapi32 and pchannel

gentoo90:
This is a rebase of #650, which was unfortunately closed by its author, with addition of `pchannel`, which is also needed for work with RDP.


leofidus:
Looks good to me. We are now using this PR successfully in our internal fork (for WTS functionality like WTSQuerySessionInformationW, I can't comment on pchannel)

tance77:
Would love to see these features in the next version.

PR #796, opened 2019-08-08
 Update audioclient & add audiopolicy

trmcnvn:
This PR updates `um::audioclient` to add missing elements and adds a new `um::audiopolicy` feature.

Closes an older PR #645

PR #868, opened 2020-02-18
add IAudioClient2+3

kcking:
Thanks for the amazing crate!
This PR adds IAudioClient2+3 which allow for lower-latency (down to 128 frame buffer size) while staying in shared mode. I have this working locally with a fork of cpal.

https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nn-audioclient-iaudioclient3 (requires Windows 10)
https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nn-audioclient-iaudioclient2 (requires Windows 8)

Do things need to be `cfg`'ed out based on feature/windows version?

### DIFF
--- a/src/um/audioclient.rs
+++ b/src/um/audioclient.rs
@@ -7,11 +7,11 @@
 use ctypes::c_float;
 use shared::basetsd::{UINT32, UINT64};
 use shared::guiddef::{LPCGUID, REFIID};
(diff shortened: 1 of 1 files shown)


kcking:
just realized https://github.com/retep998/winapi-rs/pull/796 looks like a more complete PR, but leaving this open as my vote to consider merging these new audio features :)

PR #883, opened 2020-05-03
Add definitions from oleacc.h

DataTriny:
Hello,

This PR adds all definitions from `oleacc.h`, as well as a new `oleacc` feature.

This is a requirement for UI Automation support, for which I already started the translation.

Things I am not really confident with are the function pointers on top of the file.


mwcampbell:
The function pointer type declarations at the top of oleacc.rs (using the `FN!` macro) look OK to me, but then I'm new to Rust.

I notice a couple of problems with the `pub fn` declarations below:

1. They should be enclosed in `extern "system"`, not `extern "C"`, since they use the stdcall calling convention, not cdecl.
2. The functions that are declared using `STDAPI` in the C header file return an `HRESULT`; oleacc.rs currently declares no return types for those.

That's what I've found so far.

Issue #885, opened 2020-05-11
winapi 0.4 tracking issue


retep998:
* [ ] Upgrade minimum Rust version to 1.33
* [ ] Add `#[repr(packed(N))]` everywhere applicable.
* [ ] Add `#[repr(align(N))]` everywhere applicable.
* [ ] Use actual unions for unions.
* [ ] Use nested imports.
* [ ] Upgrade to 2018 edition.
* [ ] Merge any pull requests that have breaking changes.
* [ ] Switch to [`FlexibleArray`](https://gist.github.com/rust-play/a0faca71829d3af39da4910ac9a11e30).
  * Due to needing to preserve the distinction between `[T;0]` and `[T; 1]`, it may be better to just tell people to use `VariableSizedBox`.
* [ ] Fix any deprecation warnings.
* [ ] Replace macro functions with actual const functions.
* [ ] Figure out better way to define strings.
* [ ] Get rustfmt working maybe

euclio:
Would you accept PRs tackling these issues against the `dev` branch?

retep998:
The `dev` branch is where I write my own commits for testing on CI before merging them into a mainline branch. I'll eventually create a `0.4` branch where 0.4 specific changes can be directed.

PR #887, opened 2020-05-12
Make uuidof() return a static reference

MSxDOS:
These methods currently return GUID by value which causes its unnecessary initialization in the stack instead of referring the value stored in a read-only section e.g. `.rdata`.

https://rust.godbolt.org/z/7bA-yK - a rough example of what's going on.

This is a breaking change but it shouldn't be something drastically hard to follow. The reference may always be dereferenced and then mutably referenced in rare cases where the actual value or a mutable reference is needed since GUID is `Copy`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,10 +62,10 @@ pub mod ctypes {
 // This trait should be implemented for all COM interfaces
 pub trait Interface {
     // Returns the IID of the Interface
-    fn uuidof() -> shared::guiddef::GUID;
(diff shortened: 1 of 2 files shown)


retep998:
This fails on Rust 1.6.0 and is a breaking change anyway, so it will be deferred until winapi 0.4

MSxDOS:
Yeah, the plan for this PR was to be a part of https://github.com/retep998/winapi-rs/issues/885 "pull requests that have breaking changes", should have mentioned that probably.

PR #897, opened 2020-05-27
Explicitly specify targets to build on docs.rs

jyn514:
This makes sure that all Windows platforms will have documentation on
docs.rs even after docs.rs stops building all tier 1 targets by default.
See https://github.com/rust-lang/docs.rs/pull/532 and
https://github.com/rust-lang/docs.rs/issues/343 for more context. There will be an blog post to blog.rust-lang.org/ shortly with more details (but the change is a ways in the future).

Additionally, this stops building targets on Mac and Linux since those
targets are just empty anyway.


retep998:
Handled in https://github.com/retep998/winapi-rs/commit/0090d411766ff22a2f280fc42e6a61e04780cdd4

PR #898, opened 2020-05-28
Export macro UNION

brmmm3:


retep998:
People shouldn't need to use this macro as Rust has stable unions now, which winapi will also be switching to in 0.4

PR #904, opened 2020-06-09
fix winspool printer devmode definition

dotnal:
pretty much what it says on the tin
the printer_info field for PRINTER_INFO_9W was incorrectly using a non-wide character set for pDevMode


retep998 (approved):

retep998:
Thanks for catching this!

PR #906, opened 2020-06-11
Added missing FileVersionInfo functionality

Sorenon:
Added missing functions to winver.rs
Reordered winver.rs to match the order of the original header
Replaced explicit pointers in winver.rs with their concise winapi equivalents

Created versrc.rs from [versrc.h](https://docs.microsoft.com/en-us/windows/win32/api/verrsrc/) and populated it with the constants and struct type


Cargo.toml
```diff
@@ -364,6 +364,7 @@ userenv = []
 usp10 = []
 utilapiset = []
 uxtheme = []
+versrc = []
```
roblabla:
```suggestion
verrsrc = []
```

Looks like this typo happened everywhere.

build.rs
```diff
@@ -344,6 +344,7 @@ const DATA: &'static [(&'static str, &'static [&'static str], &'static [&'static
     ("usp10", &["minwindef", "ntdef", "windef", "winerror", "wingdi", "winnt"], &["usp10"]),
     ("utilapiset", &["minwindef", "ntdef"], &["kernel32"]),
     ("uxtheme", &["commctrl", "minwindef", "windef", "wingdi", "winnt"], &["uxtheme"]),
+    ("versrc", &["minwindef", "winnt", "winuser"], &[]),
```
roblabla:
```suggestion
    ("verrsrc", &["minwindef", "winnt", "winuser"], &[]),
```

src/um/mod.rs
```diff
@@ -249,6 +249,7 @@ pub mod gl;
 #[cfg(feature = "usp10")] pub mod usp10;
 #[cfg(feature = "utilapiset")] pub mod utilapiset;
 #[cfg(feature = "uxtheme")] pub mod uxtheme;
+#[cfg(feature = "versrc")] pub mod versrc;
```
roblabla:
```suggestion
#[cfg(feature = "verrsrc")] pub mod verrsrc;
```

src/um/versrc.rs
```diff
@@ -0,0 +1,105 @@
...
+// All files in the project carrying such notice may not be copied, modified, or distributed
+// except according to those terms
+//! Version Resource definitions
+use shared::minwindef::DWORD;
+use um::winnt::LPWSTR;
+use um::winuser::RT_VERSION;
+pub const VS_FILE_INFO: LPWSTR = RT_VERSION;
+pub const VS_VERSION_INFO: DWORD = 1;
+pub const VS_USER_DEFINED: DWORD = 100;
+pub const VS_FFI_SIGNATURE: DWORD = 0xBD04EFFE;
```
roblabla:
```suggestion
pub const VS_FFI_SIGNATURE: DWORD = 0xFEEF04BD;
```

You took the _MAC version!

src/um/versrc.rs
```diff
@@ -0,0 +1,105 @@
+// Licensed under the Apache License, Version 2.0
```
roblabla: File should be renamed `verrsrc.rs`

roblabla (changes requested):
Needs a few fixes.

- Organization of code
  - All definitions go into the source file that directly maps to the header the definition is from:
    - :warning: versrc doesn't match header file name `verrsrc.h`.
  - Definitions are defined in the same order as they are in the original header: :heavy_check_mark: 

- Extern functions
  - The calling convention specified should be the one for 32bit: :heavy_check_mark: 

- Constants
  - Names are correct :heavy_check_mark: 
  - Values are correct :warning: one small issue.
  - Types are correct: All the constants are meant to be used in the VS_FIXEDFILEINFO, which is all DWORD: :heavy_check_mark: 

- Structs
  - Wrapped in the STRUCT! macro, matches C structure: :heavy_check_mark:

roblabla (approved):
LGTM :+1:

All problems from previous review fixed!

PR #910, opened 2020-06-18
Add projectedfslib

fanzeyi:
This PR adds bindings to [Projected File System](https://docs.microsoft.com/en-us/windows/win32/projfs/projected-file-system).

I want to open a pull request to avoid duplicate works. I think most things should work. However there are a few structs are not passing the structs x64 test because of possible incorrect definition of anonymous unions. 

I also want to spend some time replicating the official [RegFS example](https://github.com/microsoft/Windows-classic-samples/tree/master/Samples/ProjectedFileSystem) with this API, so we can be confident that this binding works fine.

- [x] Fix struct-x64 tests
- [x] Replicate RegFS example


src/um/projectedfslib.rs
```diff
@@ -52,7 +52,7 @@ STRUCT!{struct PRJ_EXTENDED_INFO_u_Symlink {
     TargetName: PCWSTR,
 }}
 UNION!{union PRJ_EXTENDED_INFO_u {
-    [u32; 1],
+    [u32; 1] [u64; 1],
```
retep998:
```suggestion
    [usize; 1],
```

fanzeyi:
Alright, I've finished rewriting the RegFS example in Rust. It is available here: https://github.com/fanzeyi/prjfs-rs and it uses fairly amount of the API I added in this PR. It works!

src/um/projectedfslib.rs
```diff
@@ -0,0 +1,324 @@
...
+        failureReason: *mut PRJ_UPDATE_FAILURE_CAUSES,
+    ) -> HRESULT;
+    pub fn PrjWriteFileData(
+        namespaceVirtualizationContext: PRJ_NAMESPACE_VIRTUALIZATION_CONTEXT,
+        dataStreamId: *const GUID,
+        buffer: *mut c_void,
+        byteOffset: UINT64,
+        length: UINT32,
+    ) -> HRESULT;
+    pub fn PrjGetOnDiskFileState(destinationFileName: PCWSTR, fileState: *mut PRJ_FILE_STATE);
```
xavierd: This function should return an HRESULT.

PR #933, opened 2020-08-02
Add PROC_THREAD_ATTRIBUTE_* stuff

MikailBag:


src/um/winbase.rs
```diff
@@ -361,6 +361,115 @@ pub const MS_CTS_ON: DWORD = 0x0010;
...
+#endif
+#if (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
+#define PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES \
+    ProcThreadAttributeValue (ProcThreadAttributeSecurityCapabilities, FALSE, TRUE, FALSE)
+#endif
+#define PROC_THREAD_ATTRIBUTE_PROTECTION_LEVEL \
+    ProcThreadAttributeValue (ProcThreadAttributeProtectionLevel, FALSE, TRUE, FALSE)
+#if (_WIN32_WINNT >= _WIN32_WINNT_WINBLUE)
+#endif
+*/
```
nico-abram: I don't think leaving the original header code in a comment like this is necessary, it's not specified in the guidelines and I don't see it anywhere else in the file

src/um/winbase.rs
```diff
@@ -361,6 +361,115 @@ pub const MS_CTS_ON: DWORD = 0x0010;
...
+pub const ProcThreadAttributeUmsThread: DWORD = 6;
+pub const ProcThreadAttributeMitigationPolicy: DWORD = 7;
+pub const ProcThreadAttributeSecurityCapabilities: DWORD = 9;
+pub const ProcThreadAttributeProtectionLevel: DWORD = 11;
+pub const ProcThreadAttributeJobList: DWORD = 13;
+pub const ProcThreadAttributeChildProcessPolicy: DWORD = 14;
+pub const ProcThreadAttributeAllApplicationPackagesPolicy: DWORD = 15;
+pub const ProcThreadAttributeWin32kFilter: DWORD = 16;
+pub const ProcThreadAttributeSafeOpenPromptOriginClaim: DWORD = 17;
+pub const ProcThreadAttributeDesktopAppPolicy: DWORD = 18;
```
nico-abram: Is there a reason you didn't add `pub const ProcThreadAttributePseudoConsole: DWORD = 22;`? (Unless that's new and wasn't there when you opened this PR)

nico-abram: I *think* you can use the ENUM! macro for these, and explicitly set the values like [this](https://github.com/retep998/winapi-rs/blob/a6dd5510212a2f0f5aae66e60949bd6fdaa0a44d/src/um/d2d1effects.rs#L153)

I'm not sure if that's actually better since if you do this you'll need to cast it when used in the macro

nico-abram:
I don't think this adds the things in the correct place (Within the file). Shouldn't this go right after https://github.com/retep998/winapi-rs/blob/191a475056e90d1183afe8d371b5d12cd43f479a/src/um/winbase.rs#L1248 ? (And before GetStartupInfoA)

src/um/winbase.rs
```diff
@@ -361,6 +361,115 @@ pub const MS_CTS_ON: DWORD = 0x0010;
...
+    PROC_THREAD_ATTRIBUTE_HANDLE_LIST = ProcThreadAttributeHandleList: 0,1,0;
+    PROC_THREAD_ATTRIBUTE_GROUP_AFFINITY = ProcThreadAttributeGroupAffinity: 1,1,0;
+    PROC_THREAD_ATTRIBUTE_PREFERRED_NODE = ProcThreadAttributePreferredNode: 0,1,0;
+    PROC_THREAD_ATTRIBUTE_IDEAL_PROCESSOR = ProcThreadAttributeIdealProcessor: 1,1,0;
+    PROC_THREAD_ATTRIBUTE_UMS_THREAD = ProcThreadAttributeUmsThread: 1,1,0;
+    PROC_THREAD_ATTRIBUTE_MITIGATION_POLICY = ProcThreadAttributeMitigationPolicy: 0,1,0;
+    PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES = ProcThreadAttributeSecurityCapabilities: 0,1,0;
+    PROC_THREAD_ATTRIBUTE_PROTECTION_LEVEL = ProcThreadAttributeProtectionLevel: 0,1,0;
+}
+// TODO: more attributes
```
nico-abram: I'm not sure if this TODO is needed, unless it was meant for yourself before merging?

src/um/winbase.rs
```diff
@@ -361,6 +361,115 @@ pub const MS_CTS_ON: DWORD = 0x0010;
...
+pub const ProcThreadAttributeMitigationPolicy: DWORD = 7;
+pub const ProcThreadAttributeSecurityCapabilities: DWORD = 9;
+pub const ProcThreadAttributeProtectionLevel: DWORD = 11;
+pub const ProcThreadAttributeJobList: DWORD = 13;
+pub const ProcThreadAttributeChildProcessPolicy: DWORD = 14;
+pub const ProcThreadAttributeAllApplicationPackagesPolicy: DWORD = 15;
+pub const ProcThreadAttributeWin32kFilter: DWORD = 16;
+pub const ProcThreadAttributeSafeOpenPromptOriginClaim: DWORD = 17;
+pub const ProcThreadAttributeDesktopAppPolicy: DWORD = 18;
+// TODO: constants for mitigation policy
```
nico-abram: I'm not sure if this TODO is needed, unless it was meant for yourself before merging?

src/um/winbase.rs
MikailBag: yes, this TODO is for me

src/um/winbase.rs
MikailBag: this TODO is for me too

src/um/winbase.rs
MikailBag: I do not see this define in my winbase.h
The possible cause it that I mostly use linux, and probably didn't update windows sdk for a long time (I used `10.0.17134.0`)

MikailBag:
@nico-abram Thank you for your review!

I cleaned up all temporary files, moved new definitions to better location and added all definitions present in my local windows sdk headers.

As I see, those headers are outdated a bit. I think new definitions can be added in other PR.

Regarding `ENUM!`, I am currently feeling pretty confused about types (should they all be u32 or u64) and would like to resolve this concern first.

MikailBag:
Update: `UpdateProcThreadAttribute` takes attribute as `usize` aka `DWORD_PTR`. Does it really mean all attributes declared as `ui64` simply overflow on win x86?

PR #944, opened 2020-09-22
WebAuthN bindings

hecrj:
Adds bindings to part of [the Windows WebAuthN API](https://github.com/microsoft/webauthn).

Specifically, I have only added the bindings necessary to use the `WebAuthNAuthenticatorGetAssertion` function.

This is my first contribution here and also my first time writing Rust bindings. Thus, it's very likely that I missed something!


hecrj:
I have also introduced the bindings necessary to use `WebAuthNAuthenticatorMakeCredential` in 939d8f5915aa02f4dd6d982977837361841334a0.

MaulingMonkey:
webauthn.h implies several of these structs might gain additional fields in the future (and in fact has comments implying this happened a few times already before `WEBAUTHN_API_VERSION_1` as found in `10.0.18362.0`!).  In modern Rust, `#[non_exhaustive]` might be appropriate?  But that's not available until well after winapi's minimum supported rust version of 1.6.  I'm not sure if extending the `STRUCT!` macro to support adding an old style hidden `_non_exhaustive: ()` field might be appropriate...?

`WEBAUTHN_API_VERSION_2` (`10.0.19041.0`, which this PR appears to be based on) doesn't add any new fields to any of the structs if the header comments are to be believed, so that's a plus at least.

src/um/webauthn.rs
```diff
@@ -0,0 +1,221 @@
...
+use shared::guiddef::GUID;
+use shared::minwindef::{BOOL, DWORD, PBYTE};
+use shared::windef::HWND;
+use um::winnt::{HRESULT, LONG, LPCWSTR, PCWSTR, PVOID};
+pub const WEBAUTHN_API_VERSION_1: DWORD = 1;
+pub const WEBAUTHN_API_VERSION_2: DWORD = 2;
+pub const WEBAUTHN_API_CURRENT_VERSION: DWORD = WEBAUTHN_API_VERSION_2;
+pub const WEBAUTHN_HASH_ALGORITHM_SHA_256: &'static str = "SHA-256";
+pub const WEBAUTHN_HASH_ALGORITHM_SHA_384: &'static str = "SHA-384";
+pub const WEBAUTHN_HASH_ALGORITHM_SHA_512: &'static str = "SHA-512";
```
MaulingMonkey: Nit:  `WEBAUTHN_HASH_ALGORITHM_SHA_*` doesn't match header order

src/um/webauthn.rs
```diff
@@ -0,0 +1,221 @@
...
+    pCredentials: PWEBAUTHN_CREDENTIAL,
+}}
+pub type PWEBAUTHN_CREDENTIALS = *mut WEBAUTHN_CREDENTIALS;
+pub type PCWEBAUTHN_CREDENTIALS = *const WEBAUTHN_CREDENTIALS;
+pub const WEBAUTHN_CTAP_TRANSPORT_USB: DWORD = 0x01;
+pub const WEBAUTHN_CTAP_TRANSPORT_NFC: DWORD = 0x02;
+pub const WEBAUTHN_CTAP_TRANSPORT_BLE: DWORD = 0x04;
+pub const WEBAUTHN_CTAP_TRANSPORT_TEST: DWORD = 0x08;
+pub const WEBAUTHN_CTAP_TRANSPORT_INTERNAL: DWORD = 0x10;
+pub const WEBAUTHN_CTAP_TRANSPORT_FLAGS_MASK: DWORD = 0x1F;
```
MaulingMonkey: Should these be 0-padded like they are in the header ala GUIDs?  (values are correct)

src/um/webauthn.rs
```diff
@@ -0,0 +1,221 @@
...
+    pvExtension: PVOID,
+}}
+pub type PWEBAUTHN_EXTENSION = *mut WEBAUTHN_EXTENSION;
+pub type PCWEBAUTHN_EXTENSION = *const WEBAUTHN_EXTENSION;
+STRUCT!{struct WEBAUTHN_EXTENSIONS {
+    cExtensions: DWORD,
+    pExtensions: PWEBAUTHN_EXTENSION,
+}}
+pub type PWEBAUTHN_EXTENSIONS = *mut WEBAUTHN_EXTENSIONS;
+pub type PCWEBAUTHN_EXTENSIONS = *const WEBAUTHN_EXTENSIONS;
```
MaulingMonkey, line 110: Nit:  `[P[C]]WEBAUTHN_EXTENSION[S]` placement doesn't match header order vs `WEBAUTHN_CREDENTIAL_LIST`

src/um/webauthn.rs
```diff
@@ -0,0 +1,221 @@
...
+pub type PWEBAUTHN_ASSERTION = *mut WEBAUTHN_ASSERTION;
+pub type PCWEBAUTHN_ASSERTION = *const WEBAUTHN_ASSERTION;
+extern "system" {
+    pub fn WebAuthNAuthenticatorMakeCredential(
+        hWnd: HWND,
+        pRpInformation: PCWEBAUTHN_RP_ENTITY_INFORMATION,
+        pUserInformation: PCWEBAUTHN_USER_ENTITY_INFORMATION,
+        pPubKeyCredParams: PCWEBAUTHN_COSE_CREDENTIAL_PARAMETERS,
+        pWebAuthNClientData: PCWEBAUTHN_CLIENT_DATA,
+        pWebAuthNMakeCredentialOPtions: PCWEBAUTHN_AUTHENTICATOR_MAKE_CREDENTIAL_OPTIONS,
```
MaulingMonkey: Nit: pWebAuthNMakeCredentialO**P**tions miscapitalized

src/um/webauthn.rs
```diff
@@ -0,0 +1,221 @@
...
+        ppWebAuthNCredentialAttestattion: *mut PWEBAUTHN_CREDENTIAL_ATTESTATION,
```
MaulingMonkey: Nit: ppWebAuthNCredentialAttestat**t**ion is one too many `t`s

MaulingMonkey:
> Thus, it's very likely that I missed something!

For what it's worth, these are pretty spot on as far as I'm concerned!  Good stuff!  👍

    - [ ] Definitions are defined in the same order as they are in the original header.
        ❌ A couple of minor deviations (called out in source annotations above)
    - [ ] Names are correct (functions, parameters)
        ⚠️ A couple of minor parameter name typos (only important if we care about doc comments?)
- [Constants](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#constants)
    - [x] Names are correct
    - [x] Values are correct
    - [x] Types are correct
- [Structs](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#structs)
    - [x] Each field must be on its own line.
    - [x] Struct and field names match header.
    - [x] Field types match header.
    - [ ] Misc
        ⚠️ The previously mentioned `#[non_exhaustive]` concerns.

src/um/webauthn.rs
hecrj: Should be fixed in 11662566dd9a2305be58503e2927cce8dbf42636.

src/um/webauthn.rs
hecrj: Should be fixed in 11662566dd9a2305be58503e2927cce8dbf42636.

src/um/webauthn.rs
hecrj: Changed in fc4d782e4118431d2e2517237f0b3d8704f70e38.

src/um/webauthn.rs
hecrj: Fixed in 7dc2d8502925678e89c51c5779ff2868bbfcba7c.

src/um/webauthn.rs
hecrj: Fixed in 7dc2d8502925678e89c51c5779ff2868bbfcba7c.

hecrj:
Thanks for taking a look @MaulingMonkey!

I think I took care of most of the issues. Let me know if I should rebase and squash all the commits together.

>  I'm not sure if extending the STRUCT! macro to support adding an old style hidden _non_exhaustive: () field might be appropriate...?

I am not very familiar with macros nor the internals of the library, so I don't think I can discuss this :sweat_smile: Maybe we should wait until @retep998 gets a chance to review this. No rush!

roblabla:
> webauthn.h implies several of these structs might gain additional fields in the future (and in fact has comments implying this happened a few times already before WEBAUTHN_API_VERSION_1 as found in 10.0.18362.0!). In modern Rust, #[non_exhaustive] might be appropriate?

I believe there is a *great* deal of structs in winapi that have had fields added to them as time went on. I don't know what the best way to deal with that is, but whatever it is, it should probably be punted to a 0.4 release.

PR #948, opened 2020-10-16
Never return from ExitProcess

Plecra:


MaulingMonkey:
I'm not sure if this would technically be a breaking change in some strange edge cases involving Fn type inference?  Would it be appropriate to target `dev` instead of `0.3`?

That said, the windows headers *do* mark it `DECLSPEC_NORETURN` (`__declspec(noreturn)`) which means it [really shouldn't ever return](https://docs.microsoft.com/en-us/cpp/cpp/noreturn?view=msvc-160) (would be undefined behavior in C++ codebases if it did), so I think the change itself is good :+1:

PR #949, opened 2020-10-29
Instead of an empty crate, give a compile error when not on Windows

jyn514:
Previously, programs trying to use winapi on other targets would see
errors like this:

```
failed to resolve: could not find um in winapi
... many other related errors ...
```

Now, they'll see errors like this:

```
  --- stderr
  error: winapi is only supported on Windows platforms
  help: if your application is multi-platform, use `[target.'cfg(windows)'.dependencies] winapi = "..."`
  help: if your application is only supported on Windows, use `--target x86_64-pc-windows-gnu` or some other windows platform
```

This is a breaking change. The reason for this change is two-fold:

1. For programs intending to use winapi, it gives them a much better
error message telling them how to fix the error.
2. For programs that don't need winapi, it forces them to remove it from
dependencies, decreasing their compile times.


jyn514:
CI is failing because it's still building on linux even though I told it not to, I think because it's using the CI file from the 0.3 branch instead of my PR.

MaulingMonkey:
This would be a breaking change - I believe this should target the `dev` branch instead of the `0.3` branch?

jyn514:
@retep998 what are your thoughts on this? I'm ok with closing it if you don't want it, but I do think it's a good change.

PR #952, opened 2020-11-09
Add newdev bindings

### LINKED ISSUE #851: Missing newdev
https://docs.microsoft.com/en-us/windows/win32/api/newdev/

roblabla:
Fixes #851 .

These bindings are only missing `DiInstallDevice`, which has the following signature:

```
BOOL
WINAPI
DiInstallDevice(
    _In_opt_  HWND hwndParent,
    _In_      HDEVINFO DeviceInfoSet,
    _In_      PSP_DEVINFO_DATA DeviceInfoData,
    _In_opt_  PSP_DRVINFO_DATA DriverInfoData,
    _In_      DWORD Flags,
    _Out_opt_ PBOOL NeedReboot
    );
```

However, PSP_DRVINFO_DATA is a unicode-dependent data-structure. Either the header is buggy, or the function somehow doesn't access any of the UNICODE/ASCII fields.

Since I don't need this function and couldn't find any documentation about which structure is the one we're supposed to pass, I left it out.


MaulingMonkey (approved):
LGTM! :+1:

        - `C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\newdev.h`
    - [x] Names are correct (functions, parameters)
- [Constants](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#constants)
    - [x] Names are correct
    - [x] Values are correct
    - [x] Types are correct

PR #956, opened 2020-11-12
Add missing struct in winsvc.h

nxtn:


nico-abram:
Relevant header code:

```
//
// Service delayed autostart info setting
//
typedef struct _SERVICE_DELAYED_AUTO_START_INFO {
    BOOL       fDelayedAutostart;      // Delayed autostart flag
} SERVICE_DELAYED_AUTO_START_INFO, *LPSERVICE_DELAYED_AUTO_START_INFO;
```

LGTM, the only thing is maybe we should also add the pointer type below

    pub type LPSERVICE_DELAYED_AUTO_START_INFO = *mut SERVICE_DELAYED_AUTO_START_INFO;

I also ran ripgrep in the entire include folder looking for the struct name to check for [duplicate definitions](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#dealing-with-duplicates) just in case, did not find any

MaulingMonkey (commented):
LGTM! :+1:

        - `C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winsvc.h`
- [Structs](https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#structs)
    - [x] Each field must be on its own line.
    - [x] Struct and field names match header.
    - [x] Field types match header.

PR #957, opened 2020-11-13
Add some functions from Msi and MsiQuery

roblabla:
Adds some functions from Msi.h and MsiQuery.h, useful when writing [MSI Custom Actions](https://docs.microsoft.com/en-us/windows/win32/msi/custom-action-type-1).

Those bindings were written using the 10.0.19041.0 headers.


MaulingMonkey (approved):
LGTM! :+1:

        - `C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\Msi.h`
        - `C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\MsiQuery.h`
    - [x] Names are correct (functions, parameters)
- Enums
    - [x] Enum and enumerant names are correct
    - [x] Enum values are correct
- Simple `type`s
    - [x] Names are correct
    - [x] Types are correct

Issue #958, opened 2020-11-13
Missing structure FILE_ZERO_DATA_INFORMATION in winioctl.h


RealXuChe:
`FILE_ZERO_DATA_INFORMATION` In `winioctl.h`:
```c++
typedef struct _FILE_ZERO_DATA_INFORMATION {
  LARGE_INTEGER FileOffset;
  LARGE_INTEGER BeyondFinalZero;
} FILE_ZERO_DATA_INFORMATION, *PFILE_ZERO_DATA_INFORMATION;
```
[MS Doc](https://docs.microsoft.com/en-us/windows/win32/api/winioctl/ns-winioctl-file_zero_data_information)

PR #959, opened 2020-11-14
Added um::winioctl::FILE_ZERO_DATA_INFORMATION struct.

RealXuChe:
#958


MaulingMonkey (approved):
LGTM?

* [x] The bindings match the headers (`C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winioctl.h`)
* [x] The bindings respect https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#structs
* [ ] https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md#organization-of-code
  ⚠️ The file isn't in header order, so this new definition can't be either.

PR #968, opened 2020-12-23
add definitions for afunix.h

fanzeyi:
Since Windows 17063, there is UNIX domain socket support in Windows. Along with the support `afunix.h` is added to define `SOCKADDR_UN` struct for creating UNIX domain sockets.

This pull request adds the definition mirrors `afunix.h` in Windows SDK.

See "[AF_UNIX comes to Windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)"

Windows header file location: `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\afunix.h`

### DIFF
--- a/src/shared/afunix.rs
+++ b/src/shared/afunix.rs
@@ -1,0 +1,10 @@
+use shared::minwindef::DWORD;
+use shared::ws2def::{ADDRESS_FAMILY, IOC_VENDOR};
+use um::winnt::CHAR;
+pub const UNIX_PATH_MAX: usize = 108;
+STRUCT!{struct SOCKADDR_UN {
+    sun_family: ADDRESS_FAMILY,
+    sun_path: [CHAR; UNIX_PATH_MAX],
+}}
+pub type PSOCKETADDR_UN = *mut SOCKADDR_UN;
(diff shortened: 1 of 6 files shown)


Thomasdezeeuw:
Any updates on this? We need this in socket2: https://github.com/rust-lang/socket2/pull/249.

fanzeyi:
Not sure what I can do to get this land. `winapi-rs` seems to be completely stagnated now. :(

There was some discussion about moving winapi-rs to the nursery but nothing has happened it seems.

src/shared/afunix.rs
```diff
@@ -0,0 +1,10 @@
+use shared::minwindef::DWORD;
```
elmarco, line 1: nice! I came up with practically the same patch without knowing you add a MR.

I think you should add the common top header:
```
// Licensed under the Apache License, Version 2.0
// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
// All files in the project carrying such notice may not be copied, modified, or distributed
// except according to those terms.
//! Definitions for the AF_UNIX socket address family.
```

PR #970, opened 2020-12-30
Add bindings for Debug Interface Access SDK

Osspial:


Osspial:
~~I've marked this as a draft since it still needs the class IDs.~~ they were there, just not where I was initially looking

PR #1003, opened 2021-09-03
Use core c_void type and remove need for libc

GuillaumeGomez:
It created an issue in [sysinfo](https://github.com/GuillaumeGomez/sysinfo/issues/570).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ debug = ["impl-debug"]
 everything = []
 impl-debug = []
 impl-default = []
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Oh, that requires to upgrade the minimum supported rust version...

@retep998: Should I automatically detect the version of the rustc compiler like `libc` does to enable it through a feature maybe?

retep998:
This change will be done in 0.4 when the MSRV is bumped

Issue #1014, opened 2021-12-11
License not included


BastianZim:
Hi, when downloading `winapi-x86_64-pc-windows-gnu:0.4.0` and `winapi-i686-pc-windows-gnu:0.4.0` no license files are included. Not sure why because they are specified in the `Cargo.toml`.

Ref:
* https://github.com/retep998/winapi-rs/blob/5b1829956ef645f3c2f8236ba18bb198ca4c2468/x86_64/Cargo.toml#L9
* https://github.com/retep998/winapi-rs/blob/5b1829956ef645f3c2f8236ba18bb198ca4c2468/i686/Cargo.toml#L9

ankane:
It looks like they were added on June 7, 2018, which is after the 0.4.0 release (on January 17, 2018).

BastianZim:
Ahh yes that makes sense – thanks.

PR #1019, opened 2021-12-28
feat(LICENSE): include the license files

sassman:
this small change makes sure the LICENSE files are included in the crate artefact. 

Right now they are missing. I suspect the leading `/` is not correct.


ankane:
Hey @sassman, the license files should be included in the next release without any changes - there just hasn't been a release in a while.

Ref: #623 #1014

